PR #62, opened 2024-04-17
contract variables called on every render. 

### LINKED ISSUE #47: contract variables called on every render.
## Issue Overview

Contract variables located on the left side of the `/debug` will appear as loading on every render.
- keep the  `watch : true`
- solve similar to [this issue](https://github.com/Quantum3-Labs/scaffold-stark-2/issues/39)

None

edulanasca:
# Contract variables called on every render. 

Fixes #47

## Types of change

- [ ] Feature
- [x] Bug
- [ ] Enhancement

## Comments (optional)
We have to treat `useContractRead` as a fetch function, every time `form` was changing made args to change, executing the hook several times.
I fixed by changing another state `inputValue` when was needed (inside the onClick) to prevent this behavior.

### DIFF
--- a/packages/nextjs/app/debug/_components/contract/DisplayVariable.tsx
+++ b/packages/nextjs/app/debug/_components/contract/DisplayVariable.tsx
@@ -27,6 +27,7 @@ export const DisplayVariable = ({
 }: DisplayVariableProps) => {
   const {
     data: result,
+    isLoading,
     isFetching,
     refetch,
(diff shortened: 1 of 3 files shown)


jrcarlos2000:
What has to be changed is  `contractVariables.tsx` file, as stated in the issue , that is the left side of the `debug` page . @edulanasca

0xquantum3labs (approved):

PR #76, opened 2024-04-25
Feat: Include deployer address during deployment process

### LINKED ISSUE #75: Add deployer_address during deployment contract process
## Issue Overview

For challege-0, we have to pass the `owner` input, currently we are doing this manually for devnet and testnet. https://github.com/Quantum3-Labs/speedrunstark/blob/1a8996a31f0090a66fb78bb5f47538fdf4642f26/packages/snfoundry/scripts_js/deploy.js#L19

Since this owner address will be the deployer address, the deployment script should identify the deployer address from the .env variables.

Use case: This deployer address might be different from user/caller address. For devnet our deployer address could the last devnet account and our user/caller address might be the first account. For testnet our deployer address could an argent account and our user/caller address might be a braavos account

Import .env values and automatically setup deployer address

gianalarcon:
# Task name here

Fixes #75 

## Types of change

- [x] Feature
- [x] Enhancement


packages/snfoundry/scripts_js/deploy.js
```diff
@@ -1,4 +1,10 @@
 const { deployContract } = require("./deploy_contract");
+const argv = require("yargs/yargs")(process.argv.slice(2)).argv;
+const networks = require("./helpers/networks");
+const networkName = argv.network;
+const { deployer } = networks[networkName];
+const last_devnet_account = "0x4b3f4ba8c00a02b66142a4b1dd41a4dfab4f92650922a3280977b0f03c75ee1";
+const deployer_address = deployer.address || last_devnet_account
```
jrcarlos2000, line 7: We don't need all this lines, we need to export deployer from `deploy_contract.js` and import it here similar as `deploy_contract`

jrcarlos2000 (changes requested):

PR #89, opened 2024-04-26
Parse Felt252

### LINKED ISSUE #86: Felt252 is been parsed to base 10 number
## Issue Overview

Currently a Felt252 data type is been parsed to base 10 number. This is not the desired result, we expect it to keep it(or parsed it) as hexadecimal value

Founded on challenge-1
Screenshot from 2024-04-26 23-24-41

Here `name` and `symbol` should show hexadecimal values

Fix parsing process

edulanasca:
# Parse Felt252

Fixes #86

## Types of change

- [ ] Feature
- [x] Bug
- [ ] Enhancement


packages/nextjs/utils/scaffold-stark/contract.ts
0xquantum3labs, line 1: this should go in `utils/commons.ts` , this is a helper function for an specific use case so its not included in the main `scaffold-stark` utils.

lets also use this way, is simpler

```javascript
export function feltToHex(feltBigInt: bigint) {
  return `0x${feltBigInt.toString(16)}`;
}
```

0xquantum3labs (changes requested):

PR #117, opened 2024-05-07
useScaffoldEventHistory hook

### LINKED ISSUE #107: useScaffoldEventHistory hook
## Issue Overview

Translate the following [Hook](https://github.com/scaffold-eth/scaffold-eth-2/blob/main/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts) using the starknet stack:

the process looks like this :
1. type checking and completion for evens based on a contract name through ABI ( similar to other hooks )
2. gets a public client ( see useScaffoldContractInfo )
3. reads events from the public client

To see how event reading works in starknet : [here](https://www.starknetjs.com/docs/guides/events/)
`keys` should represent the hash of the event name and any other indexed values.

for the keys use the following [pattern](https://www.starknetjs.com/docs/api/namespaces/selector/#getselectorfromname)  :
```javascript
      import { hash } from "starknet";
      events: [
        {
          keys: [hash.getSelectorFromName("Transfer")],
        },
      ],
```

edulanasca:
# useScaffoldEventHistory hook

Fixes #107

## Types of change

- [x] Feature
- [ ] Bug
- [ ] Enhancement

## Comments (optional)
- `useStarkProfile` hook is commented in `Address.tsx` out as it causes issues at `yarn build`.
```
TypeError: Cannot read properties of undefined (reading 'multicall')
```


packages/nextjs/hooks/scaffold-stark/useScaffoldEventHistory.ts
```diff
@@ -0,0 +1,299 @@
...
+            to_block: { block_number: blockNumber },
+          })
+        ).events;
+        setFromBlockUpdated(BigInt(blockNumber + 1));
+
+        const newEvents = [];
+        for (let i = logs.length - 1; i >= 0; i--) {
+          newEvents.push({
+            log: logs[i],
+            // args: logs[i].args,
```
0xquantum3labs, line 113: lets remove this comment.

packages/nextjs/hooks/scaffold-stark/useScaffoldEventHistory.ts
```diff
@@ -0,0 +1,299 @@
...
+            block:
+              blockData && logs[i].block_hash === null
+                ? null
+                : await publicClient.getBlockWithTxHashes(logs[i].block_hash),
+            transaction:
+              transactionData && logs[i].transaction_hash !== null
+                ? await publicClient.getBlockWithTxHashes(
```
0xquantum3labs, line 120: typo here we should use getTransactionByHash instead.

0xquantum3labs (changes requested):

0xquantum3labs (approved):

PR #119, opened 2024-05-10
utilsDisplay

### LINKED ISSUE #118: use Param parsers from utils globally
## Issue Overview

Inside [`utilsDisplay`](https://github.com/Quantum3-Labs/scaffold-stark-2/blob/main/packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx), we are using the starknet js utils directly to parse params.
 
Use instead the already existing ParseParamWithType from [`scaffold-stark/utils`](https://github.com/Quantum3-Labs/scaffold-stark-2/blob/main/packages/nextjs/utils/scaffold-stark/contract.ts)

Find a clean way to plug that in:
- locate where this function is used : mainly in `/debugContracts`

edulanasca:
# utilsDisplay

Fixes #118

## Types of change

- [ ] Feature
- [ ] Bug
- [x] Enhancement


packages/nextjs/utils/scaffold-stark/types.ts
```diff
@@ -0,0 +1,11 @@
...
+  ContractAddress: "core::starknet::contract_address::ContractAddress",
+  u8: "core::integer::u8",
+  u16: "core::integer::u16",
+  u32: "core::integer::u32",
+  u128: "core::integer::u128",
+  u256: "core::integer::u256",
+  usize: "core::integer::u32", // is equivalent
+  felt252: "core::felt252",
+  ByteArray: "core::byte_array::ByteArray",
+};
```
jrcarlos2000, line 11: lets try to use `Kanabi` abi types for this instead.

jrcarlos2000 (changes requested):

packages/nextjs/utils/scaffold-stark/contract.ts
```diff
@@ -388,9 +394,9 @@ export function parseParamWithType(
...
 ) {
   if (isRead) {
-    if (paramType.includes(CairoTypes.u256)) {
+    if (isCairoU256(paramType)) {
       return tryParsingParamReturnObject(uint256.uint256ToBN, param);
-    } else if (paramType.includes(CairoTypes.ByteArray)) {
+    } else if (isCairoByteArray(paramType)) {
       if (typeof param === "string") {
         return tryParsingParamReturnObject(
           byteArray.byteArrayFromString,
```
0xquantum3labs, line 402: we use `byteArrayFromString()` when writing, this is `isRead == true ` part of the code. is there any reason why we are using this here, we already had this function below in line 426. @edulanasca

0xquantum3labs (changes requested):

packages/nextjs/utils/scaffold-stark/contract.ts
0xquantum3labs: likewise for bool

```javascript
else if (isCairoBool(paramType)) {
      return tryParsingParamReturnObject(String, param);
 }
```

0xquantum3labs: we are missing `bytes31` , it should return the hex version of the number. this is only for read.
``` javascript
    } else if (isCairoBytes31(paramType)) {
      const fn = (x: bigint) => {
        return `0x${x.toString(16)}`;
      };
      return tryParsingParamReturnObject(fn, param);
    }
```

jrcarlos2000 (approved):

0xquantum3labs (approved):

PR #169, opened 2024-06-10
feat: integrate starknetID

### LINKED ISSUE #101: Resolve Starkname
## Issue Overview

1. Make sure the `Address/` component is behaving correctly with `core:::ContractAddress`, ([docs to Address component](https://docs.scaffoldeth.io/components/AddressInput))
2. allow users to pass a name for example `nadai.stark`, similar to ethereum `vitalik.eth` and the address should resolve in the same way done [here](https://docs.scaffoldeth.io/components/AddressInput) 
- Test scaffold-eth interaction and see how it works.
- find a similar solution on our framework
- use this 2 functions or 1 of them: [useStarkAddress](https://starknet-react.com/hooks/query/usestarkaddress), [useStarkName](https://starknet-react.com/hooks/query/usestarkname)
- to test resolution of addresses use [Voyager](https://voyager.online/contract/)

Marchand-Nicolas:
# Starknet Profiles

Fixes #101 

## Types of change

- [ ] Feature
- [x] Bug
- [x] Enhancement

## Comments (optional)

Also adds pfps
image


gianalarcon:
Hi @Marchand-Nicolas. Thanks for your contribution. Could you fix the PR to pass the CI? Thanks a lot.

Currently I facing this error when I try to test this PR with sepolia

image

I am not sure if `starknet.id` supports sepolia or not

Marchand-Nicolas:
> Hi @Marchand-Nicolas. Thanks for your contribution. Could you fix the PR to pass the CI? Thanks a lot.
> 
> Currently I facing this error when I try to test this PR with sepolia
> 
> image
> 
> I am not sure if `starknet.id` supports sepolia or not
> 
> How to reproduce the error:
> 
> * yarn deploy --network sepolia
> * yarn start

Starknet.id does support Sepolia, but it looks like when I exec these commands, the app starts on devnet rather than sepolia (and of course devnet is not supported). I am going to check why it doesn't use the right network

Marchand-Nicolas:
In fact the network is set in a static way for the front, so it's always devnet, that's why there is a "Network not supported" error.
image
I am going to ask StarknetID's team for a solution.

0xquantum3labs:
> In fact the network is set in a static way for the front, so it's always devnet, that's why there is a "Network not supported" error. image I am going to ask StarknetID's team for a solution.

 It would be great to either support pulling data from sepolia when `scaffold.config.ts` points to devnet, or have this functionality available ONLY when we are on sepolia or mainnet with some conditionals on `Address.tsx` @Marchand-Nicolas

gianalarcon:
> > In fact the network is set in a static way for the front, so it's always devnet, that's why there is a "Network not supported" error. image I am going to ask StarknetID's team for a solution.
> 
> It would be great to either support pulling data from sepolia when `scaffold.config.ts` points to devnet, or have this functionality available ONLY when we are on sepolia or mainnet with some conditionals on `Address.tsx` @Marchand-Nicolas

Yes. It is working well on sepolia, but it breaks when we try to use devnet. I suggest the second option "have this functionality available ONLY when we are on sepolia or mainnet with some conditionals on Address.tsx".

gianalarcon (approved):
LGTM!

PR #202, opened 2024-07-21
fix: big int input and display bug

metalboyrick:
#  Big int input and display bug

Fixes #relevant-issue-here

## Types of change

- [ ] Feature
- [X] Bug
- [ ] Enhancement

### Input Bug
- Input on fields that would parse into BigInt, (e.g. any u256 field)
- Click "Read"
- Should show "Do not know how to serialize BigInt"

### Display Bug
- Do any operations what would display in numbers outside the max or min boundary for Javascrtip Number
- Should display number incorrectly with decimal places)

## Comments (optional)

- Reference to why flatten to string: https://starknet-react.com/hooks/query/usecontractread

### DIFF
--- a/packages/nextjs/app/debug/_components/contract/ReadOnlyFunctionForm.tsx
+++ b/packages/nextjs/app/debug/_components/contract/ReadOnlyFunctionForm.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState, useRef } from "react";
+import { useState, useRef, useEffect } from "react";
 import { Abi } from "abi-wan-kanabi";
 import { Address } from "@starknet-react/chains";
 import {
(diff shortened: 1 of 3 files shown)


packages/nextjs/app/debug/_components/contract/ReadOnlyFunctionForm.tsx
```diff
@@ -36,7 +36,9 @@ export const ReadOnlyFunctionForm = ({
     address: contractAddress,
     functionName: abiFunction.name,
     abi: [...abi],
-    args: inputValue ? inputValue.flat() : [],
+    args: inputValue
```
metalboyrick, line 39: reference: https://starknet-react.com/hooks/query/usecontractread

jrcarlos2000 (approved):

Issue #194, opened 2024-07-16
Migrate `snfoundry` package to fully support Starknet foundry scripts rather than ts-scripts


gianalarcon:
## Issue Overview

Currently our declare and deploy process uses starknet.js(typescript)

## Proposed Solutions or Ideas

Migrate the declare and deploy process to cairo, you have to use sn-foundry
Use scarb 2.6.4
Use latest version of starknet-devnet
Test your solution on Sepolia

Optional: Use [v2](https://github.com/Quantum3-Labs/scaffold-stark-2/tree/v2)

Your solution must no break the current features and functionalities of the framework.

### Instructions for Contributing

- Leave a comment on this Issue.
- Wait for our confirmation to start on the project.
- Once you have received confirmation, begin working on your contribution.

### Incentives

Contributions are rewarded via Only Dust.

### Contact

For questions, join to our [tg group](https://t.me/+wO3PtlRAreo4MDI9) and reach out our team

PR #223, opened 2024-07-30
Migrated Snfoundry package to support starknet foundry script

Akinbola247:
Migrate snfoundry package to fully support Starknet foundry scripts rather than ts-scripts #194

BREAKING CHANGES

Fixes #relevant-issue-here

## Types of change
- [x] Enhancement

## Comments (optional)
ADDED ENHANCEMENT :
1. Reverted to starknet devnet 0.0.4 to match with main
2. Added cairo script to declare and deploy
3. (Yarn deploy automated)Added deploy command to declare and deploy using cairo script. `yarn deploy ` or `yarn deploy --network sepolia`
4. Deployer Account info is auto generated to snfoundry.toml in the process of running `yarn deploy`
5. Modified parse-deployment.ts script
6.  Added native Sncast in command.ts to automate deployment and network specification
7. Racfactored folder structure & renamed scripts.

TO DO
- When class hash already exists or If the class already exists just deploy (PENDING DUE TO SNFOUNDRY BUG).

BUG
1. Compile time with snfoundry takes a little while compared to starknetjs
2. <bug from snfoundry> When the declare scripts runs the second time, it is bound to fail because the classhash has already been declared which is meant to return a byteArray as an error message. unfortunately, the msg: ByteArray isn't made pub hence not visible. As result of this, starknet-devnet has to be restarted after every deploy runs locally, and for sepolia/mainnet, already declared classhash has to be specified and passed in manually in the script.cairo


0xquantum3labs:
Hey @Akinbola247 🙌🏾 Thanks for your contribution this is having a better look now. 

Some TODOs :
- Declare fails on a second run on same network
    - when class already exists
    - If the class already exists just deploy ( not sure how possible is this ) 
    - The autogenerated script.json file is not updated after repeated deploy runs, this is why its failing on NEXTJS because its using a contract deployed on a different instance of devnet which doesnt exist anymore
        - Quick solution delete the file and rerun.
        - Maybe a bug ?
        - we shouldnt commit this run-script.json files , can we pass the folder where these are written to through a cli param like `--output` ?

- Find a way to automate `yarn deploy` without having to add an account , at least on localhost. Maybe generate a scaffold-default account that uses the first private key of localhost chain. 
    - I should be able to do `yarn chain`, `yarn deploy`. And have my contracts ready on UI.
- Try to AVOID as much as we can commands like yarn `deploy:devnet` , we should do `yarn deploy  - -network {devnet | sepolia | mainnet}`
- starknet-devnet-rs version 0.0.4, make sure everything is in same line, as `main` branch uses this version,
- Currently there is `snfoundry-dev/contracts` . Let’s remove this nested folders and pass everything inside `contracts/` to be on `snfoundry/ package.json` and `scarb.toml` & ``snfoundry.toml`` will be on the same folder.

0xquantum3labs:
hi @Akinbola247 , let me know when the TODOs are fixed and your branch is updated, and conflicts are resolved  💯

Akinbola247:
Oh yes, i will. just one last todo. currently fixing

Akinbola247:
due to a bug in sncast_std library, this todo :" When class hash already exists or If the class already exists just deploy" is yet to be completed. When the declare scripts runs the second time, it is bound to fail because the classhash has already been declared which is meant to return a byteArray as an error message. unfortunately, the msg: ByteArray isn't made pub  hence not visible. All things being equal, every other Todos has been completed.

jrcarlos2000:
Hello @Akinbola247 , im getting this error when running 

`yarn deploy --network sepolia`

notes :
- I updated my contract source such that the class hash is a new class hash
- i have updated my `.env`
- works well on devnet

image

lets also remove this type of messages that i get when running the script, try to be close to the current messages on `main` branch:

```
command: script run
status: success
command: account add

```

packages/snfoundry/scripts-ts/helpers/command.ts
```diff
@@ -0,0 +1,89 @@
...
+
+function generateNameString(length: number): string {
+  return crypto.randomBytes(length / 2).toString("hex");
+}
+
+function deploy(network: string = "devnet"): void {
+  const deployerName = generateNameString(8);
+
+  let command: string;
+  if (network === "sepolia") {
```
jrcarlos2000, line 31: missing "mainnet"

jrcarlos2000 (changes requested):

Akinbola247:
Alright!, i'll push a fix in a bit. Thanks.

Akinbola247:
`command: script run
status: success
command: account add` 
 
The above message cannot be removed, as it is native to sncast

0xquantum3labs (approved):

jrcarlos2000 (approved):

PR #226, opened 2024-08-01
feat: add support for STRK balance

raizo07:
# Add STRK token balance

Fixes #https://github.com/Quantum3-Labs/scaffold-stark-2/issues/212

## Types of change

- [x] Feature


packages/nextjs/contracts/deployedContracts.ts
0xquantum3labs, line 1: this file cant change

0xquantum3labs (changes requested):

0xquantum3labs:
this seems clean and working well, just a few things to point out:
1. We can generalize the an ERC20 impl that is the same as the current `Eth` ABI.

currently STRK fails for some functions like `increaseAllowance`, i get `entrypoint not found`, this is probably an due to a wrong ABI.

we can use one single ABI for Eth and STRK since the usage of this is mostly for ERC20 transactions ( i.e. transfers, approves, transferFrom )

```typescript
const universalERC20Abi = ......   // previous Eth Abi

eth = {
   address,
   abi : universalERC20ABi
   ....
}
Strk - {
   address,
   abi = universalERC20ABI
}
```

2. When clicking on balance i get 2 amounts in $ . 
image

instead i should be able to get a sum of the value of both token. i.e. 3000$ worth of ETH + 1000$ worth of STRK 
i should visualize 4000$ when clicking.

packages/nextjs/contracts/predeployedContracts.ts
0xquantum3labs, line 1: normalize the use of one single ERC20 Abi, rather than multiple ABI implementations for each token in each network.

0xquantum3labs (changes requested):

packages/nextjs/contracts/predeployedContracts.ts
```diff
@@ -2,13 +2,219 @@
  * This file is autogenerated by Scaffold-Stark.
  * You should not edit it manually or your changes might be overwritten.
  */
-const universalEthAddress =
+const universalERC20Abi =
```
0xquantum3labs, line 5: why are we replacing EthAddress & StrkAddres with Abi ?

packages/nextjs/components/scaffold-stark/Balance.tsx
```diff
@@ -21,6 +22,9 @@ export const Balance = ({ address, className = "", usdMode }: BalanceProps) => {
   const { formatted, isLoading, isError } = useScaffoldEthBalance({
     address,
   });
+  const strkBalance = useScaffoldStrkBalance({
```
0xquantum3labs, line 25: can we follow the same pattern as before , so the code looks cleaner. 

```typescript
const { formatted, isLoading, isError } = useScaffoldEthBalance()
```

we can use different names like `EthBalanceFormatted`,`StrkBalanceFormatted`.

packages/nextjs/components/scaffold-stark/Balance.tsx
```diff
@@ -52,33 +62,49 @@ export const Balance = ({ address, className = "", usdMode }: BalanceProps) => {
     );
   }
 
-  //const formattedBalance = balance ? Number(balance.formatted) : 0;
+  // Calculate the total balance in USD
+  const ethBalanceInUsd = parseFloat(formatted) * price;
+  const strkBalanceInUsd = parseFloat(strkBalance.formatted) * price;
```
0xquantum3labs, line 67: why are we using the same price for Eth and Strk ?

0xquantum3labs (changes requested):

0xquantum3labs:
@raizo07 please check the failing workflow

packages/nextjs/components/scaffold-stark/Balance.tsx
raizo07: > why are we using the same price for Eth and Strk ?

@0xquantum3labs  From here you can see that the comment on the review asked for the  sum of the value of both tokens

https://github.com/Quantum3-Labs/scaffold-stark-2/pull/226#issuecomment-2264788700

image

0xquantum3labs: the value of ETH and STRK are not the same @raizo07

0xquantum3labs: the price of STRK and the price of ETH are not the same
@raizo07

0xquantum3labs: @raizo07 gm any updates ?

raizo07: Hello @0xquantum3labs  the targetNetwork here always returns Eth.

How do I get it to return Strk

image

0xquantum3labs: you can create another hook `useStrkCurrencyPrice()` and adapt the  `fetchPriceFromCoinGecko()` function

packages/nextjs/utils/scaffold-stark/fetchStrkPriceFromCoingecko.ts
jrcarlos2000, line 1: Can we just modify the original function 'fetchPriceFromCoinGecko()' in order to receive a **token symbol** as parameter rather than a **ChainWithAttributes**

@raizo07

jrcarlos2000 (changes requested):

0xquantum3labs (approved):

jrcarlos2000 (approved):

PR #220, opened 2024-07-29
chore: imporved deployement script

ikemHood:
# Improve deploy function on deploy-contract.ts

Fixes #217 

utilized maxFee
removed unused imports
imporved logging
tested on devnet and sepolia

- [ ] Feature
- [ ] Bug
- [✅ ] Enhancement

PR #231, opened 2024-08-03
Improve deploy function on deploy-contract.ts

### LINKED ISSUE #217: Improve `deploy` function on `deploy-contract.ts`
## Issue Overview

- update the order of functions, able to receive null constructor params
- the only `required` param is the contractName.
- receives the contract name first, then optional `exportContractName`, then the optional`constructor params`,  and finally the `options`
- makes good usage of  `options` and `maxFee`
- remove unused imports
- good error logging for all cases !
- other dev improvements ++

- should be tested on all networks.

ikemHood:
Fixes #217 
Improvements on #220 

✅ No file changes in packages/nextjs
✅ added colorized logging

### DIFF
--- a/packages/snfoundry/scripts-ts/deploy-contract.ts
+++ b/packages/snfoundry/scripts-ts/deploy-contract.ts
@@ -3,40 +3,40 @@ import path from "path";
 import { networks } from "./helpers/networks";
 import yargs from "yargs";
 import {
-  BlockIdentifier,
   CallData,
-  hash,
   stark,
   RawArgs,
-  constants,
-  ec,
-  validateAndParseAddress,
   transaction,
-} from "starknet";
-import { Network } from "./types";
-import {
-  LegacyContractClass,
-  CompiledSierra,
   extractContractHashes,
+  DeclareContractPayload,
+  UniversalDetails,
 } from "starknet";
+import { DeployContractParams, Network } from "./types";
 
 const argv = yargs(process.argv.slice(2)).argv;
 const networkName: string = argv["network"];
 
 let deployments = {};
-
 let deployCalls = [];
+let deploymentOptions: UniversalDetails;
 
 const { provider, deployer }: Network = networks[networkName];
 
-const declareIfNot_NotWait = async (payload: any) => {
+const declareIfNot_NotWait = async (
+  payload: DeclareContractPayload,
+) => {
   const declareContractPayload = extractContractHashes(payload);
   try {
     await provider.getClassByHash(declareContractPayload.classHash);
   } catch (error) {
-    let { transaction_hash } = await deployer.declare(payload);
-    if (networkName == "sepolia" || networkName == "mainnet") {
-      await provider.waitForTransaction(transaction_hash);
+    try {
+      const { transaction_hash } = await deployer.declare(payload, deploymentOptions);
+      if (networkName === "sepolia" || networkName === "mainnet") {
+        await provider.waitForTransaction(transaction_hash);
+      }
+    } catch (e) {
+      console.error("Error declaring contract:", e);
+      throw e;
     }
   }
   return {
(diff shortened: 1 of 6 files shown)


ikemHood:
@0xquantum3labs Please review

packages/snfoundry/scripts-ts/deploy-contract.ts
```diff
@@ -145,14 +184,14 @@ const deployContract = async (
     constructorCalldata,
   });
 
-  console.log("Contract Deployed at ", contractAddress);
+  console.log(green("Contract Deployed at "), contractAddress);
 
-  let finalContractName = exportContractName || contractName;
+  let finalContractName = contract || contractName;
```
0xquantum3labs, line 189: the `contractName` variable comes first, if its set, it shoulld take this one as it is the exportName indicated by the user.

packages/snfoundry/scripts-ts/deploy.ts
```diff
@@ -4,21 +4,64 @@ import {
...
+ * };
+ *
+ *
+ * @returns {Promise<void>}
+ */
 const deployScript = async (): Promise<void> => {
-  await deployContract(
-    {
-      owner: deployer.address, // the deployer address is the owner of the contract
+  await deployContract({
```
0xquantum3labs, line 45: lets keep this to the bare minimum, just `deploy(contractName, constructor )`, we already have examples above.

0xquantum3labs (changes requested):

0xquantum3labs:
when i run 

```typescript
  await deployContract({
    contract: "YourContrac1t",
    constructorArgs: {
      owner: deployer.address,
    },
  });
```

if the contract is not found, it should fail and dont send transactions, however i get this : 

image

packages/snfoundry/scripts-ts/deploy-contract.ts
```diff
@@ -163,17 +202,21 @@ const deployContract = async (
 
 const executeDeployCalls = async () => {
```
0xquantum3labs, line 203: missing optional maxFee as options. this is a maxFee that will be taken into account for only this function and shouldnt be overwritten on each call to `deployContract`. This one is a different parameter from the one in `declareIfNot_NotWait`

packages/snfoundry/scripts-ts/deploy-contract.ts
```diff
@@ -49,46 +51,73 @@ const deployContract_NotWait = async (payload: {
...
-  options?: {
-    maxFee: bigint;
-  }
+  params: DeployContractParams
 ): Promise<{
   classHash: string;
   address: string;
 }> => {
+  const { contract, constructorArgs, contractName, options } = params;
+  deploymentOptions = options;
```
0xquantum3labs, line 96: this `options.maxFee` is meant to be passed to `declareIfNot_NotWait` as declareOptions. 
this current implementation keeps track of one single `deploymentOptions` which is updated on each call, if this options are not present on a call it will use the previous one.

0xquantum3labs (changes requested):

packages/snfoundry/scripts-ts/deploy-contract.ts
ikemHood: if no options was passed on a new call, options is undefined and overwrites the previously set options.

The reason i am doin it this way is to allow the user set the options for `declareIfNot_NotWait` and `executeDeployCalls`  once, largely because of of how we call the script.

```ts
deployScript()
  .then(async () => {
    await executeDeployCalls();
    exportDeployments();

    console.log(green("All Setup Done"));
  })
  .catch(console.error);
```

but since you have specified that they are different values, i would go ahead and allow users set them independently

0xquantum3labs (approved):

PR #246, opened 2024-08-11
Fix: Clean files

gianalarcon:
# Task name here

In order to make a smooth merge process form scaffold to speedrunstark, suggest to Clean files

## Types of change

- [x] Enhancement


packages/nextjs/components/scaffold-stark/Faucet.tsx
```diff
@@ -41,7 +41,7 @@ export const Faucet = () => {
   useEffect(() => {
     const checkChain = async () => {
       try {
-        const providerInfo = await publicClient.getBlock();
+        const providerInfo = await publicClient.getBlockWithTxHashes();
```
0xquantum3labs, line 44: it seems to me that this will be a heavier request, which we are not supposed to do when calling `checkChain()` only, why do we need `txHashes` ?

packages/nextjs/contracts/deployedContracts.ts
0xquantum3labs, line 1: we cant commit this file

0xquantum3labs (changes requested):

gianalarcon:
Fixed! There are some `consolge.log` you guys can check if we really need them or not.

packages/nextjs/app/debug/page.tsx
```diff
@@ -12,16 +12,7 @@ const Debug: NextPage = () => {
...
-        <h1 className="text-4xl my-0">Debug Contracts</h1>
-        <p className="text-neutral">
-          You can debug & interact with your deployed contracts here.
-          <br /> Check{" "}
-          <code className="italic bg-base-300 text-base font-bold [word-spacing:-0.5rem] px-1">
-            packages / nextjs / app / debug / page.tsx
-          </code>{" "}
-        </p>
-      </div> */}
+      {}
```
metalboyrick, line 15: can remove this bracket altogether i think

metalboyrick (changes requested):

0xquantum3labs (approved):

metalboyrick (approved):
lgtm

PR #249, opened 2024-08-12
fix: clicking twice to open read

### LINKED ISSUE #244: Bug on "Read" button on debug contracts
## Issue Overview

On first load of a contract UI inside `/debug` page, when reading a function, need to press 2 times on button READ
image

- needs to work on first interaction. 
- Use and test the `enabled` flag on the `useContractRead` hook of Starknet-React.
- a previous issue was the `enabled` flag  not working on useContractRead of Starknet-react. but that seems to have been solved

metalboyrick:
# Task name here

Fixes #244 

## Types of change

- [ ] Feature
- [X] Bug
- [ ] Enhancement

## Comments (optional)

- Also optimized the callback `handleRead` with the `useCallback` hook

### DIFF
--- a/packages/nextjs/app/debug/_components/contract/ReadOnlyFunctionForm.tsx
+++ b/packages/nextjs/app/debug/_components/contract/ReadOnlyFunctionForm.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState, useRef, useEffect } from "react";
+import { useState, useRef, useEffect, useCallback } from "react";
 import { Abi } from "abi-wan-kanabi";
 import { Address } from "@starknet-react/chains";
 import {
@@ -38,7 +38,7 @@ export const ReadOnlyFunctionForm = ({
     functionName: abiFunction.name,
     abi: [...abi],
     args: inputValue ? inputValue.flat(Infinity) : [],
-    enabled: false,
+
     parseArgs: false,
     blockIdentifier: "pending" as BlockNumber,
   });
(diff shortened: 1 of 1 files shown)


packages/nextjs/app/debug/_components/contract/ReadOnlyFunctionForm.tsx
```diff
@@ -38,7 +38,7 @@ export const ReadOnlyFunctionForm = ({
     functionName: abiFunction.name,
     abi: [...abi],
     args: inputValue ? inputValue.flat(Infinity) : [],
-    enabled: false,
```
0xquantum3labs, line 41: is this set by default to `false` on starknet-react ? if not, the idea of having this is that we dont execute many `FAILED` calls on first render , for example would affect if we have longer contracts or RPCs with limited bandwidth. any way we can get around this ?

metalboyrick: Default is `true`. the thing is that this is a wrapper around  `react-query`'s `useQuery`. The supposed behaviour is that to lazy load, this flag is set to `false`. However, for some reason in here requires calling `refetch` twice. 

One alternative would be to use input value as enabled flag? that way it's true only when input value is available, this way we can skip the `FAILED` calls

discussion by react-query team: https://github.com/TanStack/query/discussions/1205#discussioncomment-109277

0xquantum3labs: yes that sounds like a good approach, avoiding the FAILED calls

metalboyrick:
@0xquantum3labs please help to reproduce and validate the failed calls ser, thank you! I can't reproduce them, so cannot really validate my code is working or not, thank you!

0xquantum3labs:
> @0xquantum3labs please help to reproduce and validate the failed calls ser, thank you! I can't reproduce them, so cannot really validate my code is working or not, thank you!

on the previous commit [5340c78](https://github.com/Quantum3-Labs/scaffold-stark-2/pull/249/commits/5340c78b14fd9fdd09ef27e1fad18f1da906710d) :

you can try 

```typescript
const { isFetching, data, refetch, error } = ....
```

do a lazy load, wait a few seconds and all the errors will be visible

All is working well with the last commit [fd69e94](https://github.com/Quantum3-Labs/scaffold-stark-2/pull/249/commits/fd69e940a7b45ef3385a1ca529f5889147d81a54).

@metalboyrick

metalboyrick:
validated, should be good to go

0xquantum3labs (approved):

PR #250, opened 2024-08-13
Update useContractRead hook

gianalarcon:
- I faced issues using useContractRead hook, then I updated this way. Please check what do you think @metalboyrick @jrcarlos2000 

- Remove this files to avoid conflict during sync of scaffold-stark and speedrun/challenges

## Types of change

- [x] Enhancement

### DIFF
--- a/packages/nextjs/hooks/scaffold-stark/useScaffoldReadContract.ts
+++ b/packages/nextjs/hooks/scaffold-stark/useScaffoldReadContract.ts
@@ -28,15 +28,11 @@ export const useScaffoldReadContract = <
     address: deployedContract?.address,
     abi: deployedContract?.abi,
     watch: true,
-    args,
-    enabled: !Array.isArray(args) || !args.some((arg) => arg === undefined),
+    args: args ?? [],
+    enabled: true,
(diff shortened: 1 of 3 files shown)


packages/nextjs/hooks/scaffold-stark/useScaffoldReadContract.ts
```diff
@@ -28,15 +28,11 @@ export const useScaffoldReadContract = <
     address: deployedContract?.address,
     abi: deployedContract?.abi,
     watch: true,
-    args,
-    enabled: !Array.isArray(args) || !args.some((arg) => arg === undefined),
+    args: args ?? [],
+    enabled: true,
```
metalboyrick, line 32: might need to be careful with enabling everything right on the start, because this will immediately fetch the contract. my suggestion is to default to `true` and then allow the user to override? wdyt @0xquantum3labs

0xquantum3labs: I agree with you, we are able to `override` this flag right now.

0xquantum3labs (approved):

PR #253, opened 2024-08-15
refactor: use useScaffoldReadContract in ContractReadMethods

### LINKED ISSUE #229: Abstract all login on Debug Page to Scaffold Hooks
## Issue Overview

Currently Debug Page uses Starknet React hooks. Considering to use Scaffold Stark hooks instead so that we can unify and sync the functionality of the debug page and the scaffold hooks. 

- Refactor the ReadOnlyFunctionForm, ContractVariables, and WriteOnlyFunctionForm reading hooks into scaffold hooks
- Include parsing on scaffold hooks
- Introduce an optional parsing flag to give the users more freedom on parsing the input data
- Add documentation regarding feature changes

metalboyrick:
# Task name here

Fixes #229 

## Types of change

- [ ] Feature
- [ ] Bug
- [X] Enhancement

## Comments (optional)

Nice to have - use context API to prevent prop drilling on some cases

### DIFF
--- a/packages/nextjs/app/debug/_components/contract/ContractReadMethods.tsx
+++ b/packages/nextjs/app/debug/_components/contract/ContractReadMethods.tsx
@@ -11,8 +11,10 @@ import { ReadOnlyFunctionForm } from "./ReadOnlyFunctionForm";
 
 export const ContractReadMethods = ({
   deployedContractData,
+  contractName,
 }: {
   deployedContractData: Contract<ContractName>;
+  contractName: ContractName;
 }) => {
   if (!deployedContractData) {
     return null;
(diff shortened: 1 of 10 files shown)


metalboyrick:
NOTE: latest commit is breaking

metalboyrick:
@jrcarlos2000 for the write part, im a bit in doubt on whether we should use `useScaffoldWriteContract` since it involves manual parsing from string to rawArgs (this is in v2, in v3 this problem has been handled witha new parsing logic). What do you think?

metalboyrick:
So far can successfully migrate read, but struggling a bit with write. Also, do we want to handle this in v3 or v2? if v3 i will close this PR and create a new one

jrcarlos2000:
> NOTE: latest commit is breaking

i think v3 of `starknet-react` already supports some parsing of arguments, so it would be better to have this done on the v3 branch, rather than our current. Otherwise it will become a loop when we migrate to the new one.

metalboyrick:
will perform changes on v3 branch

metalboyrick:
closed due to moving changes to v3

PR #258, opened 2024-08-20
fix: add workaround on wallet connectors to remove from localStorage

### LINKED ISSUE #256: BUG: In some cases, `useAutoConnect` will keep reconnecting even if user has clicked disconnect
## Issue Overview

In some cases, `useAutoConnect` will keep reconnecting even if user has clicked disconnect. Produced in https://github.com/Quantum3-Labs/speedrunstark/issues/110.

- Delete localStorage data upon disconnecting
- (Nice-to-have) abstract away `disconnect()` function cleanly to delete the data without additional calls

metalboyrick:
# Task name here

Fixes #256 
Fixes https://github.com/Quantum3-Labs/speedrunstark/issues/110

## Types of change

- [ ] Feature
- [X] Bug
- [ ] Enhancement

## Comments (optional)

- Note this is  a workaround that tampers with the `disconnect()` function of the connector. I've injected a local storage removal operation to clear local storage so that `useAutoConnect()` does not go in a infinite loop when it finds something stored.
- Should be able to resolve the issue in speedrunstark (https://github.com/Quantum3-Labs/speedrunstark/issues/110) as well @gianalarcon

### DIFF
--- a/packages/nextjs/services/web3/connectors.tsx
+++ b/packages/nextjs/services/web3/connectors.tsx
@@ -1,4 +1,4 @@
-import { argent, braavos } from "@starknet-react/core";
+import { argent, braavos, InjectedConnector } from "@starknet-react/core";
 import { getTargetNetworks } from "~~/utils/scaffold-stark";
 import { BurnerConnector } from "./stark-burner/BurnerConnector";
 import scaffoldConfig from "~~/scaffold.config";
@@ -6,6 +6,17 @@ import scaffoldConfig from "~~/scaffold.config";
 const targetNetworks = getTargetNetworks();
 
 export const connectors = getConnectors();
+
+// workaround helper function to properly disconnect with removing local storage (prevent autoconnect infinite loop)
+function withDisconnectWrapper(connector: InjectedConnector) {
+  const connectorDisconnect = connector.disconnect;
+  const _disconnect = (): Promise<void> => {
+    localStorage.removeItem("lastUsedConnector");
+    return connectorDisconnect();
+  };
+  connector.disconnect = _disconnect.bind(connector);
+  return connector;
+}
 
 function getConnectors() {
   const { targetNetworks } = scaffoldConfig;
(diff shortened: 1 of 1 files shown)


packages/nextjs/services/web3/connectors.tsx
```diff
@@ -18,7 +29,7 @@ function getConnectors() {
     connectors.push(new BurnerConnector());
   }
 
-  return connectors.sort(() => Math.random() - 0.5);
+  return connectors.sort(() => Math.random() - 0.5).map(withDisconnectWrapper);
```
gianalarcon, line 32: It will be also a great chance to refactor this sort using Math.random() - 0.5, what do you think? @jrcarlos2000

metalboyrick: are we considering to use `lodash` ?

0xquantum3labs: @metalboyrick i think something like [this](https://github.com/Quantum3-Labs/scaffold-stark-2/pull/152/files) ?

0xquantum3labs: this connectors file was updated when we tried to have a conditional connector list, that doesnt show the burner connector when the target network is not `devnet`.

0xquantum3labs (approved):

packages/nextjs/services/web3/connectors.tsx
0xquantum3labs: if its possible to go back to use [this](https://github.com/apibara/starknet-react/blob/852f202a48e6a347ffa950fd4fb2c794702699b6/packages/core/src/connectors/discovery.ts#L21) hook from `starknet-react` , wdyt ? @metalboyrick

metalboyrick: better make new pr i think, refactoring change

metalboyrick: options: either to use lodash shuffle or implement our own array shuffles, or use this hook, i think both works. Either way we still need the disconnect wrapper to ensure that wallet disconnects AND removes the local storage entry

metalboyrick: just checked the `useInjectedConnector` from starknet react, they use `lodash` shuffle under the hood haha.

PR #259, opened 2024-08-23
feat(predployedContracts): support download Contract from address

### LINKED ISSUE #238: fetch ABI from network and add instructions to append into predeployedContracts.ts
## Issue Overview

Currently users are limited to the `predeployedContracts.ts` provided by default on the monorepo.
We want to allow users to be able to fetch the ABI and receive a prompt on how to append this Contract into their `predeployedContracts` file

After this is achieved, users will be able to download a contract and then use it in the app through the scaffold hooks or use it on the `/debug` page for testing.

Steps :
- add a page next to debug to input an address and fetch the ABI from the network
- parse the Contract with its ABI and add instructions to append into `predeployedContracts.ts`
    - the Abi format fetched from network is different from the one we have, should be able to fix this with a parser function
    - clear insttructions of what to do to append should be written, ( which file, where to append ?, which network )
- implement some good error handling as well

Hints : 
- use functions like [this one](https://www.starknetjs.com/docs/api/classes/providerinterface/#getclassat) to fetch the class of an address through the provider. The class object should contain an ABI element on it.

oceanlvr:
# Task name here

Fixes #238

## Types of change

- [x] Feature
- [ ] Bug
- [ ] Enhancement

## Comments (optional)

> Currently users are limited to the predeployedContracts.ts provided by default on the monorepo.
We want to allow users to be able to fetch the ABI and receive a prompt on how to append this Contract into their predeployedContracts file
After this is achieved, users will be able to download a contract and then use it in the app through the scaffold hooks or use 
it on the /debug page for testing.

Support downloading contracts from addresses.
 
Once the user gets his contract, he can directly replace the configuration items in the predeployedContract.ts file to use his own Contract on the debug page and test abi


packages/nextjs/app/configure/_components/DownloadContracts.tsx
```diff
@@ -0,0 +1,124 @@
...
+            <span className="text-network">{targetNetwork.name}</span>
+          </div>
+        )}
+        <div className="flex items-center my-2">
+          <div className="w-24 font-medium break-words text-function">
+            Contract:
+          </div>
+          <select
+            value={symbol}
+            onChange={(e) => setSymbol(e.target.value as "Eth" | "Strk")}
```
0xquantum3labs, line 94: We shouldnt be an option between "Eth" and "Strk" only, we should allow the user to set a Name to the downloaded contract, and this name has to be different ( ignoring case ) to all the contracts on `deployedContracts` and `preDeployedContracts`.

packages/nextjs/app/configure/_components/DownloadContracts.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const handleDownload = async () => {
+    if (!address) return;
+
+    const [apiResponse, classHash] = await Promise.all([
+      provider.getClassAt(address),
+      provider.getClassHashAt(address),
+    ]);
+
+    const contractsData = {
+      [targetNetwork.network]: {
```
0xquantum3labs, line 23: we should generate the exact `.ts` file that APPENDS the dowloaded contract to the already existing contracts on `predeployedContracts.ts` . check [this](https://github.com/Quantum3-Labs/scaffold-stark-2/blob/86026df6f9b463ae2095a01f675397044137fa41/packages/snfoundry/scripts-ts/helpers/parse-deployments.ts#L78-L85) as reference, however this example is used on the `deployedContracts.ts` . 

The `predeployedContracts.ts` file downloaded should be : { ... existing data, new_data }.

0xquantum3labs: also update instructions, no need to append, rather just replace the `predeployedContracts.ts`

0xquantum3labs: the current format in `.json` cannot easily be appended into the `.ts` file

0xquantum3labs (changes requested):

packages/nextjs/app/configure/_components/DownloadContracts.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const { targetNetwork } = useTargetNetwork();
+  const [symbol, setSymbol] = useState<"Eth" | "Strk">("Eth");
+
+  const { provider } = useProvider();
+  const [address, setAddress] = useState<string>("");
+
+  const handleDownload = async () => {
+    if (!address) return;
+
+    const [apiResponse, classHash] = await Promise.all([
```
0xquantum3labs, line 17: missing error handling

0xquantum3labs (changes requested):

packages/nextjs/app/configure/_components/DownloadContracts.tsx
oceanlvr: i think it would be a litte complex if we support update `predeployedContracts.ts` file online like use ` { ... existing data, new_data }.`.
because predeployedContracts is a typescript file not json file.
which means that we need to compile and transform  this file to javascript. and btw we'll extract all the constant from this file, which means we also need to import more than two file online. 

 so I suggest  two way:
* let the user themself to do the contract relacement
* we need a new file like `onlinePredeployedContract.ts`, and merge `onlinePredeployedContract.ts` `predeployedContracts .ts` `deployedContracts.ts` those three file in the code.

any comment is welcome :) @0xquantum3labs 


TLDR; need to transform ts file and use the compl tool like `ts-morph`(online version babel)

oceanlvr: plan 2:

https://github.com/Quantum3-Labs/scaffold-stark-2/pull/259/files#diff-3f3873a1760af995681d03ae297ac7f9f1b0349f5a7f151020905a9d3b1b0c43R104-R105

oceanlvr:
```
we just provide the user a new predeployedContracts.ts he can directly copy paste it into the existing one, without affecting the existing contracts. Then the user can refresh and it should be good to go
```

sir i have made a new file name `configDeployedContracts.ts` which is download from this page, and user can replace and refresh, and it should be ok then.

oceanlvr:
PTAL sir @0xquantum3labs

packages/nextjs/app/configure/_components/DownloadContracts.tsx
oceanlvr: sir updated, ptal~ @0xquantum3labs  feel free to review

:)

packages/nextjs/contracts/configDeployedContracts.ts
0xquantum3labs, line 1: can we rename this file and everything relevant to `configExternalContracts.ts`

packages/nextjs/app/configure/_components/DownloadContracts.tsx
```diff
@@ -0,0 +1,141 @@
...
+*/`;
+
+    const fileContent = JSON.stringify(contractsData, null, 2);
+    const preDeployedContracts = `${generatedContractComment}\n\nconst preDeployedContracts = ${fileContent} as const;\n\nexport default preDeployedContracts;`;
+
+    const blob = new Blob([preDeployedContracts], { type: "text/typescript" });
+    const url = URL.createObjectURL(blob);
+    const a = document.createElement("a");
+    a.href = url;
+    a.download = "predeployedContracts.ts";
```
0xquantum3labs, line 55: doesnt this have a different name ? not `predeployedContracts.ts` ?

packages/nextjs/app/configure/_components/DownloadContracts.tsx
```diff
@@ -0,0 +1,141 @@
...
+            <li>Enter the contract address in the input field.</li>
+            <li>Enter the contract name in the input field.</li>
+            <li>Click the Download Contract File button.</li>
+            <li>
+              The tool will fetch the ABI, address, and classHash from the
+              network and generate a configuration file.
+            </li>
+            <li>
+              Download the file and replace it to your local
+              `predeployedContracts.ts` file.
```
0xquantum3labs, line 86: isnt this the `configDeployedContracts.ts`

packages/nextjs/app/configure/_components/DownloadContracts.tsx
```diff
@@ -0,0 +1,141 @@
...
+  };
+
+  const generateContractsFile = (contractsData: Object) => {
+    const generatedContractComment = `/**
+* This file is autogenerated by Scaffold-Stark.
+* You should not edit it manually or your changes might be overwritten.
+*/`;
+
+    const fileContent = JSON.stringify(contractsData, null, 2);
+    const preDeployedContracts = `${generatedContractComment}\n\nconst preDeployedContracts = ${fileContent} as const;\n\nexport default preDeployedContracts;`;
```
0xquantum3labs, line 49: we should keep whatever is already existing on `configDeployedContracts.ts`

0xquantum3labs (changes requested):

jrcarlos2000 (approved):

packages/nextjs/app/configure/_components/DownloadContracts.tsx
jrcarlos2000, line 1: intended functionality missing

jrcarlos2000 (changes requested):

packages/nextjs/app/configure/page.tsx
```diff
@@ -0,0 +1,18 @@
+import DownloadContracts from "./_components/DownloadContracts";
+import type { NextPage } from "next";
+import { getMetadata } from "~~/utils/scaffold-stark/getMetadata";
+
+export const metadata = getMetadata({
+  title: "Configure Contracts",
```
Nadai2010, line 6: It works fine.
Captura desde 2024-08-29 09-56-35

But the title and description don't load, I see it like this.

Captura desde 2024-08-29 09-41-20

packages/nextjs/app/page.tsx
```diff
@@ -54,10 +53,10 @@ const Home: NextPage = () => {
                 height={25}
               ></Image>
               <p>
-                Tinker with your smart contract using the{" "}
+                Tinker with your smart contract using the
```
Nadai2010, line 56: Be careful with this change, check

Captura desde 2024-08-29 09-44-33
Captura desde 2024-08-29 09-44-25

Nadai2010 (changes requested):

packages/nextjs/app/page.tsx
oceanlvr: Sorry for the missing check

packages/nextjs/app/configure/page.tsx
oceanlvr: image

Sir, these codes are for `i18n` and are not directly displayed on the page. They will serve for crawler robots

Nadai2010: Ok perfect, you should adjust a title or a superior design, or at least separate the upper border.

oceanlvr: OK sir let me adjust the container padding and border margin.

oceanlvr: image

make the title bolder and adjust the padding and border margin.

0xquantum3labs:
Seems good for now, pending a better design , cc : @Nadai2010

packages/nextjs/app/configure/page.tsx
Nadai2010: Here I propose for now to keep something similar DebugContract.tsx [L27](https://github.com/Quantum3-Labs/scaffold-stark-2/blob/main/packages/nextjs/app/debug/_components/DebugContracts.tsx#L27), it would follow the line better.

Now
imagen

With changes
imagen



imagen

oceanlvr:
sir update the design. PTAL @Nadai2010  @0xquantum3labs

Nadai2010:
Hi @oceanlvr, here you have an updated version of how the component should look in both versions.

imagen

imagen

oceanlvr:
> Hi @oceanlvr, here you have an updated version of how the component should look in both versions.

 copy that, looks good~👍

Nadai2010:
In light mode it should be white

imagen

oceanlvr:
fixed sir @Nadai2010

Nadai2010 (approved):

0xquantum3labs:
functionality seems good to me @metalboyrick please take a look

metalboyrick (approved):
looks good to me so far

PR #261, opened 2024-08-23
fix: do not show wallet not deployed in wrong network

### LINKED ISSUE #260: [BUG] Wallet Not Deployed incorrect rendering
## Issue Overview

Wallet not deployed message rendered on incorrect occasions:
1.  Connected to wrong network
2. Fresh account deployed but still showing the error

Also happening: https://github.com/Quantum3-Labs/speedrunstark/issues/111

- Create a new wallet
- Try to connect , should display "wallet not deployed", this is expected since we have not deployed the wallet yet
- Fund the wallet and deploy (DO NOT REFRESH)
- Disconnect from the wallet 
- Reconnect the wallet
- Wallet not deployed should still be shown.

NOTE: Refreshing removes this bug.

- Check for "wallet not deployed" rendering logic
- Check rerendering logic, might be some cache issues

metalboyrick:
# Task name here

Fixes #260 
FIxes https://github.com/Quantum3-Labs/speedrunstark/issues/111

## Types of change

- [ ] Feature
- [X] Bug
- [ ] Enhancement

### DIFF
--- a/packages/nextjs/components/Header.tsx
+++ b/packages/nextjs/components/Header.tsx
@@ -78,18 +78,18 @@ export const Header = () => {
   const isLocalNetwork = targetNetwork.id === devnet.id;
 
   const { provider } = useProvider();
-  const { address, status } = useAccount();
+  const { address, status, chainId } = useAccount();
   const [isDeployed, setIsDeployed] = useState(true);
 
   useEffect(() => {
-    if (status === "connected" && address) {
+    if (status === "connected" && address && chainId === targetNetwork.id) {
       provider.getContractVersion(address).catch((e) => {
         if (e.toString().includes("Contract not found")) {
           setIsDeployed(false);
         }
       });
     }
-  }, [status, address, provider]);
+  }, [status, address, provider, chainId, targetNetwork.id]);
 
   return (
     <div className="sticky lg:static top-0 navbar min-h-0 flex-shrink-0 justify-between z-20 px-0 sm:px-2">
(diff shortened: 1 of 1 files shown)


packages/nextjs/components/Header.tsx
```diff
@@ -78,18 +78,23 @@ export const Header = () => {
...
   useEffect(() => {
-    if (status === "connected" && address) {
-      provider.getContractVersion(address).catch((e) => {
-        if (e.toString().includes("Contract not found")) {
-          setIsDeployed(false);
-        }
-      });
+    if (status === "connected" && address && chainId === targetNetwork.id) {
+      provider
+        .getContractVersion(address)
```
0xquantum3labs, line 87: lets change this to use `getClassHashAt` , i experienced some weird behavior using this helper function. When moving from non deployed account to another non deployed account, also this will be similar to the way we do it [here](https://github.com/Quantum3-Labs/scaffold-stark-2/blob/33111d7ab6deb94e359c30d3aa917402e0dba865/packages/nextjs/hooks/scaffold-stark/useDeployedContractInfo.ts#L41) in `useScaffoldContractInfo()`

0xquantum3labs (changes requested):

packages/nextjs/components/Header.tsx
0xquantum3labs: cc : @Nadai2010

0xquantum3labs (approved):

PR #269, opened 2024-08-30
fix issue264

### LINKED ISSUE #264: Bug: Switch to Starknet Devnet button not working
## Issue Overview

The `Switch to Starknet Devnet` button is not functioning correctly when attempting to switch UI from the Sepolia network to Starknet Devnet.

## Steps to Reproduce

1. Clone this repository.
2. Deploy the contract successfully on the Sepolia network and interact with `YourContract`.
3. Attempt to deploy again, this time targeting Starknet Devnet. To do this, edit the `scaffold.config.ts` file and set Devnet as the target network. 
   - Relevant file: [`scaffold.config.ts`](https://github.com/Quantum3-Labs/scaffold-stark-2/blob/a2974ae8d1132a8889753c9a7602cffdf20b1a9f/packages/nextjs/scaffold.config.ts#L12)
4. Check the UI opened at `http://localhost:3000`. You should see a `wrong network` message.
   - wrong network message
5. Attempt to switch to Starknet Devnet using the `Switch to Starknet Devnet` button, but the button doesn't function.
   - switch to devnet issue
6. As a workaround, you can click the `Disconnect` button, then reconnect using the Burnet wallet to switch to the Devnet network.

## Proposed Solution

Fix the bug so that the `Switch to Starknet Devnet` button works as intended. The solution should ensure compatibility with Argent X and Braavos wallets as well.

Buuh2511:
# Task name here

Fixes #264 

## Types of change

- [ ] Feature
- [ ] Bug
- [ ] Enhancement

## Comments (optional)
**1. Devnet on StarkNet is Often Based on Goerli Testnet**
  - StarkNet primarily uses two networks for testing environments:
  - Goerli Testnet (SN_GOERLI): This is a public test network for StarkNet that simulates mainnet conditions but uses test ETH from the Goerli testnet.
  - Devnet: Typically, StarkNet’s development environment (devnet) is based on Goerli. This means that if you’re using "devnet," you’re essentially operating in a Goerli-like environment with slight modifications for development purposes.

**2. Why "devnet" is Mapped to "SN_GOERLI"**
  - When developers work with StarkNet, they usually develop and test on the Goerli Testnet because it provides an environment similar to the mainnet, but without real costs and lower risks.
  - Although there could be a separate network called "devnet," StarkNet treats Goerli Testnet as the primary testing environment. Hence, in many cases, devnet is essentially a clone of Goerli with specific development tweaks.
  - This explains why "devnet" is mapped to "SN_GOERLI" instead of "SN_DEVNET". This is how StarkNet unifies testing and development environments to ensure consistency.

### DIFF
--- a/packages/nextjs/hooks/scaffold-stark/useSwitchNetwork.ts
+++ b/packages/nextjs/hooks/scaffold-stark/useSwitchNetwork.ts
@@ -1,4 +1,14 @@
 
+declare global {
+  interface Window {
+    starknet?: {
+      isConnected: boolean;
+      request: (params: { type: string; params: { chainId: string } }) => Promise<void>;
+    };
+  }
+}
+
+
 const getChainId = (network: string): string => {
   switch (network) {
     case "mainnet":
@@ -13,15 +23,11 @@ const getChainId = (network: string): string => {
 export const useSwitchNetwork = () => {
   return {
     switchNetwork: async (network: string) => {
-      // @ts-ignore
       if (window.starknet && window.starknet.isConnected) {
-        const chainId = getChainId(network);
-
-        // @ts-ignore
         await window.starknet.request({
           type: "wallet_switchStarknetChain",
           params: {
-            chainId: chainId,
+            chainId: getChainId(network),
           },
         });
       }
--- a/packages/nextjs/hooks/scaffold-stark/useSwitchNetwork.ts
+++ b/packages/nextjs/hooks/scaffold-stark/useSwitchNetwork.ts
@@ -1,15 +1,27 @@
+
+const getChainId = (network: string): string => {
+  switch (network) {
+    case "mainnet":
+      return "SN_MAIN";
+    case "devnet":
+      return "SN_GOERLI";
+    default:
+      return `SN_${network.toUpperCase()}`;
+  }
+};
+
 export const useSwitchNetwork = () => {
   return {
     switchNetwork: async (network: string) => {
       // @ts-ignore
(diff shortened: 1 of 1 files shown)


packages/nextjs/hooks/scaffold-stark/useSwitchNetwork.ts
```diff
@@ -1,15 +1,27 @@
...
+      return "SN_GOERLI";
+    default:
+      return `SN_${network.toUpperCase()}`;
+  }
+};
+
 export const useSwitchNetwork = () => {
   return {
     switchNetwork: async (network: string) => {
       // @ts-ignore
```
metalboyrick, line 16: Please help to refactor to `@ts-expect-error` so that we can enforce why we need to skip the intepreter there, thank you!

packages/nextjs/hooks/scaffold-stark/useSwitchNetwork.ts
```diff
@@ -1,15 +1,27 @@
...
+        const chainId = getChainId(network);
+
         // @ts-ignore
         await window.starknet.request({
           type: "wallet_switchStarknetChain",
           params: {
-            chainId: `SN_${
-              network == "mainnet" ? "MAIN" : network.toUpperCase()
-            }`,
+            chainId: chainId,
```
metalboyrick, line 24:
```suggestion
            chainId: getChainId(network),
```

unless `chainId` is used somewhere else.

metalboyrick (changes requested):

PR #317, opened 2024-10-09
Remove unnecessary files

Nadai2010:
# Remove unnecessary files

Delete some unnecessary or duplicate files

## Types of change

- [x] Enhancement


metalboyrick (approved):
Should be OK

PR #331, opened 2024-10-17
Solution in event issue

Nadai2010:
# Solution in event issue

In the contract.ts file in the `decodeParamWithType` we have returned to the simpler version by removing `getChecksumAddress` and adding `validateAndParseAddress`.

Example Speedrun CH-0

Captura desde 2024-10-17 22-07-43

## Types of change

- [x] Bug


metalboyrick (approved):
works and looks good to me

PR #338, opened 2024-10-31
add a new chain type only for forked mainnet

### LINKED ISSUE #313: Bug: Burner wallet not working with a fork of mainnet on devnet
## Issue Overview

-Run devnet with a fork of mainnet: `yarn chain --fork-network https://starknet-mainnet.public.blastapi.io/rpc/v0_7`
⁃Deployment works with burners, can't be problem of devnet
⁃Account class hash , signature mismatch. Find workaround to import same account type as the forked network.

 ⁃ Try native starknet js tx on deploy script with burner account on fork
 ⁃ If previous work, check transactor file and BurnerWallet files on next js app.
 ⁃ Replicate the starknet js tx using starknet react.

## Temporary solution using current version
 Using your extension wallets on devnet:
 ⁃ Run chain with fork
 ⁃ Connect to developer network on bravos extension
 ⁃ Get your account address and use faucet to get eth
 ⁃ Deploy your account through the settings of braavos wallet, as you would do with any other account on your extension
 ⁃ You are able to interact with it as if it was a devnet burner account
Outline your proposed solution or share any ideas you have for discussing or resolving the issue.

iossocket:
# Add a new chain option `forkMainnet` for `scaffoldConfig`

Fixes #313 

## Types of change

- [ ] Feature
- [ ] Bug
- [x] Enhancement

## Comments

If developer start the local testnet from a mainnet fork chain, something like `starknet-devnet --fork-network https://starknet-mainnet.g.alchemy.com/starknet/version/rpc/v0_7/PLACE_HOLDER --seed 0 --account-class cairo1`, then can change `targetNetworks` to `[forkMainnet]`

### DIFF
--- a/packages/nextjs/scaffold.config.ts
+++ b/packages/nextjs/scaffold.config.ts
@@ -9,6 +9,27 @@ export type ScaffoldConfig = {
   autoConnectTTL: number;
 };
 
+const forkMainnet = {
+  id: BigInt("0x534e5f4d41494e"),
+  network: "devnet",
+  name: "Starknet Devnet",
+  nativeCurrency: {
+    address: "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
+    name: "Ether",
+    symbol: "ETH",
+    decimals: 18
+  },
+  testnet: true,
+  rpcUrls: {
+    default: {
+      http: []
+    },
+    public: {
+      http: ["http://localhost:5050/rpc"]
+    }
+  }
+} as chains.Chain;
+
 const scaffoldConfig = {
   targetNetworks: [chains.devnet],
   // Only show the Burner Wallet when running on devnet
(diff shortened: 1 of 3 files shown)


packages/nextjs/scaffold.config.ts
jrcarlos2000, line 1: thanks for this PR, can we create an external file of `customNetworks`, this will be very helpful for other people who gave feedback on trying to use other networks that should be compatible for example `katana`. in this file we can include the example `mainnetFork`. Get all the custom networks inside `scaffold.config.ts` and override `chains = [...reactChains, ...customNetworks]`.


Would that work ? @iossocket 🤔

jrcarlos2000 (changes requested):

packages/nextjs/scaffold.config.ts
iossocket: you are right, I updated it in the following commit

jrcarlos2000 (approved):

PR #348, opened 2024-11-07
feat: add support for options and results

### LINKED ISSUE #314: Feat: Handle `Option` type as input parameter
## Issue Overview

Currently Scaffold stark `Debug` can not handle  `Option` type (Some and None) as input parameter
This is the [testing branch](https://github.com/Scaffold-Stark/scaffold-stark-2/tree/gian/sc-input-option). On this branch the tests use Option and None, and `yarn test` works well.

Explore parsing of this type

metalboyrick:
# Task name here

Fixes #314 

## Types of change

- [X] Feature
- [ ] Bug
- [ ] Enhancement

## Comments (optional)

- Supports `Option` and `Result` types.
- All enum now have Radio Selection for better input validation and coercion


metalboyrick:
converting to draft as it is still a little bit buggy + not reconciled yet

metalboyrick:
ready for review

packages/nextjs/components/scaffold-stark/Input/InputBase.tsx
```diff
@@ -62,7 +62,7 @@ export const InputBase = <
     <div className={`flex bg-input text-accent ${modifier}`}>
       {prefix}
       <input
-        className="input input-ghost focus-within:border-transparent focus:outline-none focus:bg-transparent h-[2.2rem] min-h-[2.2rem] px-4 border w-full text-xs placeholder:text-[#9596BF] text-neutral"
+        className="input input-ghost focus-within:border-transparent focus:outline-none focus:bg-transparent h-[2.2rem] min-h-[2.2rem] px-4 border w-full text-xs placeholder:text-[#9596BF] text-neutral rounded-none"
         placeholder={placeholder}
         name={name}
         value={value?.toString()}
```
iossocket, line 68: When switch between undefined and some value, it may throw warning about `controlled component` and `uncontrolled component`, for our usage, it is an controlled component, better to have a default value like this: `value?.toString() || ""`

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
```diff
@@ -76,6 +85,22 @@ export const displayTxResult = ({
...
+    }
+
+    if (displayContent instanceof CairoResult) {
+      if (displayContent.isOk()) {
+        JSON.stringify({ Ok: displayContent.unwrap() }, replacer);
+      }
+
+      return JSON.stringify({ Err: displayContent.unwrap() }, replacer);
+    }
+
```
iossocket, line 103: this part will have code conflictions with latest decoding logic, would you mind to do a code rebase and also add some unit tests to cover this part as well.

iossocket:
image

Do we support this kind of input, I tried to text raw type and `Option::Some(....)`

packages/nextjs/components/scaffold-stark/Input/InputBase.tsx
metalboyrick: will fix! thanks for pointing that out!

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
metalboyrick: got it, i was rushing when doing this PR so that some guys in the HH wanted to try it. But now that it's more lax i can start adding UTs as well, will fix!

metalboyrick:
> image
> Do we support this kind of input, I tried to text raw type and `Option::Some(....)`

supposed to be

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
metalboyrick: i just used your changes and seems to be working good, will add the UTs

metalboyrick:
@iossocket do you have this contract that i can deploy to reproduce the case?

metalboyrick:
@iossocket kindly re-review

iossocket:
image

Field for Option<ContractAddress>  and Option<u128> are different?

```
#[derive(Drop, Serde)]
struct SomeStruct {
    addr: Option<ContractAddress>,
    val: Option<u128>,
}
...
fn echo_struct(self: @ContractState, input: SomeStruct) -> SomeStruct {
    input
}
```

packages/nextjs/app/debug/_components/contract/Struct.tsx
```diff
@@ -104,15 +104,27 @@ export const Struct = ({
...
-                    stateObjectKey={key}
-                    paramType={variant}
-                  />
+                  <div key={index} className="flex items-center gap-3">
+                    <input
+                      type="checkbox"
+                      name={`radio-${index}`}
+                      className="radio radio-xs radio-secondary"
+                      checked={index === activeVariantIndex}
+                      onClick={() => {
```
iossocket, line 113: for an checkbox input the action callback should be `onChange`? or keep onClick and add an empty onChange.

we can see the error msg:
```
You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`. 
```

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
```diff
@@ -205,6 +210,12 @@ const _decodeContractResponseItem = (
     return decoded;
   }
 
+  // option and result are enums but we don't want to process them as enums
+  // we need the fallback so that it does not crash
+  if (isCairoOption(abiType.name || "") || isCairoResult(abiType.name || "")) {
```
iossocket, line 215: image
should check `abiType.type`?

packages/nextjs/app/debug/_components/contract/Struct.tsx
metalboyrick: i will go for the latter for a quicker fix

packages/nextjs/app/debug/_components/contract/utilsDisplay.tsx
metalboyrick: hmm i tried to use type but it seemed to bypass this check, name works but maybe its a coincidence with the variable name (i'm using contracts from `v3-bulletproof-contracts`)

metalboyrick: oh i know why, here is the reason, was following this ABI, i provided a workaround to catch both `name` and `type`:

image

metalboyrick:
@iossocket @Nadai2010 @jrcarlos2000  pls re review, i think if there is any tech debt we can record it as an issue, if functionality is good i think can proceed first because the PR is needed for tomorrow's event, thanks guys! and sorry for rushing

iossocket:
> image
> Field for Option and Option are different?
> 
> ```
> #[derive(Drop, Serde)]
> struct SomeStruct {
>     addr: Option<ContractAddress>,
>     val: Option<u128>,
> }
> ...
> fn echo_struct(self: @ContractState, input: SomeStruct) -> SomeStruct {
>     input
> }
> ```

@metalboyrick Hi Rich, will this be an issue for tmr's event? if not, Im ok with the pull request. :)

metalboyrick:
Could be, do you have a .cairo file for this? Sorry im not very fluent with cairo haha

metalboyrick:
@iossocket finally fixed!

iossocket (approved):

PR #350, opened 2024-11-09
Fix/configure contract

### LINKED ISSUE #349: Configure contracts format doesn't match with deployedContract
## Issue Overview

Configure contracts format doesn't match with deployedContract, due to stringified, the key became a string

Rather than only stringified, use prettier formatter

ngjupeng:
# Fix configure contract format doens't match with deployedContract 

Fixes #349 

## Types of change

- [ ] Feature
- [ ] Bug
- [X] Enhancement


packages/nextjs/app/api/configure-contract/route.ts
```diff
@@ -0,0 +1,36 @@
...
+*/`;
+
+    const fileContent = await prettier.format(
+      `${generatedContractComment}\n\nconst configExternalContracts = ${JSON.stringify(data)} as const;\n\nexport default configExternalContracts;`,
+      {
+        parser: "typescript",
+      },
+    );
+
+    fs.writeFileSync(
```
iossocket, line 23: Will it be used in `Vercel`, from my understanding, `Vercel` uses a serverless function (not a traditional server) to serve files. `fs.writeFileSync` may fail in production.

metalboyrick: @iossocket i dont think this page is supposed to go to production at all. its a helper tool for the users to debug existing deployed contracts that they didnt do themselves. But i agree with @iossocket on the usage here, was wondering why is the file generation not done in client side?

metalboyrick (changes requested):

packages/nextjs/app/api/configure-contract/route.ts
Nadai2010: This feature works very well, and facilitates automatic download and upload without having to replace the file, but as the guys @metalboyrick  and @iossocket say, I don't know if it would be optimal or if it could fail,

Could we have this version from the client side???

ngjupeng: the idea was to make user convenient without the need to replace the file, but considering production, i think we can have this feature for localhost, and file generation for production?

Nadai2010: IMO, this would be a good option.

- Automatic file upload operation localhost
- File download and replacement in production.

ngjupeng: @metalboyrick @iossocket what do you think

metalboyrick: i meant that in production, people wouldnt need this feature since the application would be used by consumer not the devs haha, devs would only need this for local or staging environments to extract contract ABIs from the network.

I was thinking, we CAN actually have server-side API (which one works then we just go lah, don't think its a big deal) , in the local, but we need to ensure (or at least have a guide on this in the docs) that the devs that will be using this feature will be able to shut it off somehow (if server), because server cost can rack up too in vercel haha, and vercel is famous for being very expensive

cc @iossocket 

p.s. pls ping me on slack or tele if I dont respond in a while ser, github notifications get very cluttered very fast haha, sorry for the late reply

metalboyrick:
no updates yet since last review, will review again when there are updates

Nadai2010:
Hi @ngjupeng  this file was optimised in this #365 , check the changes [here](https://github.com/Scaffold-Stark/scaffold-stark-2/pull/365/files#diff-de184dcf0c761a84229caa47d350a3dee8730b0f091764ffbe46b1d7e78466ea)

packages/nextjs/app/configure/_components/DownloadContracts.tsx
```diff
@@ -130,7 +130,7 @@ export default function DownloadContracts() {
               </div>
               <input
                 value={symbol}
-                onChange={handleInputChange}
+                onChange={(e) => setSymbol(e.target.value)}
```
metalboyrick, line 129: what symbol does this refer to?

metalboyrick (commented):
minor comments

packages/nextjs/package.json
```diff
@@ -31,6 +31,7 @@
     "next": "14.1.3",
     "next-themes": "^0.2.1",
     "nprogress": "^0.2.0",
+    "prettier-standalone": "^1.3.1-0",
```
metalboyrick, line 34: for the others please note new `yarn install` needed @Nadai2010 @iossocket @0xquantum3labs @jrcarlos2000

packages/nextjs/app/configure/_components/DownloadContracts.tsx
ngjupeng: contract name

metalboyrick: might want to change it haha, it will confuse other ppl if they make changes or read the codebase

metalboyrick (approved):
fix the variable naming issue and i think we're good ser

packages/nextjs/app/api/configure-contract/route.ts
0xquantum3labs: @iossocket 👀

PR #353, opened 2024-11-14
CORS policy issue when accessing coingecko in local environment & refactoring price cache logic

iossocket:
# CORS policy issue

## Types of change

- [ ] Feature
- [ ] Bug
- [x] Enhancement

## Comments (optional)
1. Fixed the issue in local environment: 
```
Access to fetch at 'https://api.coingecko.com/api/v3/simple/price?ids=starknet&vs_currencies=usd' from origin 'http://localhost:3000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```
2. Refactoring price cache logic, keep the price data in a singleton PriceService instance. The api will only be called when some component use `useNativeCurrencyPriceV2`, and the `setInterval` will be initialised once, and can be canceled when no subscriber exist.

> `useNativeCurrencyPriceV2` can replace `useNativeCurrencyPrice` later


0xquantum3labs:
why wouldnt we use V2 as default right now ? @iossocket

iossocket:
Hi @0xquantum3labs , I raised the pull request to make sure the behaviour is what we want. and it can be replaced by changing the useNativeCurrencyPrice import path.
If it is ok for us, I will append the new commit and also apply this change to scaffold-eth-stark. thanks

Nadai2010 (commented):
The useNativeCurrencyPriceV2.ts works perfectly both locally, sepolia and in mainnet, the errors, warnings and calls in loops have been fixed.

Why don't we replace it with useNativeCurrencyPrice?

iossocket:
@Nadai2010 Hi, in the following commits, `useNativeCurrencyPrice` has been replaced, thanks

Nadai2010:
Thank you for the wonderful contributions

metalboyrick (approved):
@iossocket magic fix! lgtm

PR #364, opened 2024-12-01
fix : handle edge case for deployed contracts

0xquantum3labs:
# Improve contract file finding on deploy script

I realized sometimes there will be this pattern:
`../contracts/target/dev/contracts_contracts_${contract}_${contract}.compiled_contract_class.json`

and not:

`../contracts/target/dev/contracts_${contract}.compiled_contract_class.json`

this finding solution works in a better way by finding the file matching it with its tail. 

## Types of change

- [ ] Feature
- [ ] Bug
- [x] Enhancement

### DIFF
--- a/packages/snfoundry/scripts-ts/deploy-contract.ts
+++ b/packages/snfoundry/scripts-ts/deploy-contract.ts
@@ -109,6 +109,24 @@ const deployContract_NotWait = async (payload: {
   }
 };
 
+const findContractFile = (
+  contract: string,
+  fileType: "compiled_contract_class" | "contract_class"
+): string => {
+  const targetDir = path.resolve(__dirname, "../contracts/target/dev");
+  const files = fs.readdirSync(targetDir);
+
+  // Look for files that end with the contract name and file type
+  const pattern = new RegExp(`.*${contract}\\.${fileType}\\.json$`);
+  const matchingFile = files.find((file) => pattern.test(file));
+
+  if (!matchingFile) {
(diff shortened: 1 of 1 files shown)


packages/snfoundry/scripts-ts/deploy-contract.ts
```diff
@@ -109,6 +109,24 @@ const deployContract_NotWait = async (payload: {
...
 
+const findContractFile = (
+  contract: string,
+  fileType: "compiled_contract_class" | "contract_class"
+): string => {
+  const targetDir = path.resolve(__dirname, "../contracts/target/dev");
+  const files = fs.readdirSync(targetDir);
+
+  // Look for files that end with the contract name and file type
+  const pattern = new RegExp(`.*${contract}\\.${fileType}\\.json$`);
```
metalboyrick, line 120: @jrcarlos2000 does `fileType` refer to things like `compiled_contract_class`

packages/snfoundry/scripts-ts/deploy-contract.ts
```diff
@@ -109,6 +109,24 @@ const deployContract_NotWait = async (payload: {
...
+  const matchingFile = files.find((file) => pattern.test(file));
+
+  if (!matchingFile) {
+    throw new Error(`Could not find ${fileType} file for contract ${contract}`);
```
metalboyrick, line 124: might need to leave a suggestion here on what the guy should do if they got this error

metalboyrick (changes requested):

metalboyrick (approved):
looks good from my side

PR #382, opened 2024-12-12
Extract stark burner to a separate lib

iossocket:
# Extract stark burner

leverage github insight feature, stark-burner lib: https://github.com/Scaffold-Stark/stark-burner

## Types of change

- [x] Feature
- [ ] Bug
- [ ] Enhancement


metalboyrick:
@iossocket converting to draft to make sure this PR doesnt get mixed up in the OD Hack PRs

metalboyrick (commented):
Where is the burner code stored ser?

metalboyrick (approved):
code on this side looks good, thanks for the unit tests as well

metalboyrick (commented):
might need to do a reconciliation after OD ends

packages/nextjs/services/web3/connectors.tsx
```diff
@@ -23,12 +23,13 @@ function withDisconnectWrapper(connector: InjectedConnector) {
...
 
-  const connectors = [argent(), braavos()];
+  const connectors: InjectedConnector[] = [argent(), braavos()];
 
   if (
     targetNetworks.some((network) => (network.network as string) === "devnet")
   ) {
-    connectors.push(new BurnerConnector());
+    const burnerConnector = new BurnerConnector();
+    connectors.push(burnerConnector as unknown as InjectedConnector);
```
metalboyrick, line 32: just curious what is the reason to refactor it like this?

iossocket: The root cause of this issue lies in the fact that `InjectedConnector` base class from `@starknet-react/core` has a private `_wallet` property, while `BurnerConnector` as a subclass doesn't explicitly define this property.
In TypeScript, when a class inherits from another class, if the parent class has private properties, the child class must somehow "know" about the existence of these private properties. Due to the scope restrictions of private properties, these properties must come from the same declaration context (same package/module).

metalboyrick: i see, thanks!

Nadai2010 (approved):
I have tested it, for my part it is perfect, everything works fine, very clean work, thank you very much.

0xquantum3labs (approved):

PR #392, opened 2024-12-14
fix: remove scrollbar produced when QR button is clicked

Kingsuite:


packages/nextjs/styles/globals.css
```diff
@@ -27,9 +27,28 @@ p {
   @apply shadow-md;
 }
 
+html {
+  @apply overflow-y-scroll;
+}
+
+body {
+  @apply overflow-y-auto;
+}
```
metalboyrick, line 36: can consider hiding overflow x? what do you think?

metalboyrick (changes requested):
minor comments!

packages/nextjs/styles/globals.css
Nadai2010: Hi @Kingsuite please remember the required changes

Kingsuite: GM @Nadai2010 @metalboyrick, sorry for the delay. Changes have been implemented. Thanks for your review

PR #398, opened 2024-12-18
Update Bug_0xblank

### LINKED ISSUE #345: [BUG] `0x<blank>` input considered valid on some fields.
## Issue Overview

Fields like `felt252` can accept this input and will still enable the read/send button if the input is `0x[blankl]`. Expected should be disabled button with tooltip.

image

either:
- use number input for felt
- check the current input with validation

KevinMB0220:
# Address Input Validation Fix

Fixes #345 

## Types of change

- [x] Bug

## Comments (optional)

Fixed a bug where the **AddressInput** allowed invalid inputs such as `0x`.  
Now the input is validated to ensure it meets the correct format:  
- It must start with `0x`.  
- It must contain only hexadecimal characters.

If validation fails, an error message is displayed: **"Invalid address format"**.


packages/nextjs/components/scaffold-stark/Address.tsx
```diff
@@ -96,8 +96,8 @@ export const Address = ({
     );
   }
 
-  if (!validateChecksumAddress(checkSumAddress)) {
-    return <span className="text-error">Wrong address</span>;
+  if (!checkSumAddress || !/^0x[0-9a-fA-F]+$/.test(checkSumAddress)) {
```
metalboyrick, line 99: Possible to wrap these regex into an util function?

packages/nextjs/components/scaffold-stark/Faucet.tsx
```diff
@@ -119,7 +119,11 @@ export const Faucet = () => {
             <EtherInput
               placeholder="Amount to send"
               value={sendValue}
-              onChange={(value) => setSendValue(value)}
+              onChange={(value) => {
+                if (/^0x[0-9a-fA-F]+$/.test(value)) {
```
metalboyrick, line 123: Same as here

metalboyrick (changes requested):

packages/nextjs/components/scaffold-stark/Address.tsx
KevinMB0220: I'll wrap the regex into a utility function as suggested. Thanks for pointing it out!

packages/nextjs/components/scaffold-stark/Faucet.tsx
KevinMB0220: Ok

packages/nextjs/components/scaffold-stark/Address.tsx
```diff
@@ -60,6 +60,11 @@ export const Address = ({
     checkSumAddress || address || "",
   );
 
+  const isValidHexAddress = (value: string): boolean => {
+    const hexAddressRegex = /^0x[0-9a-fA-F]+$/;
+    return hexAddressRegex.test(value);
+  };
+
```
metalboyrick, line 67: lets put it in the utils file (i think we have this util already). we don't want duplicates of the same function since they would be a pain to maintain

KevinMB0220: I was given another recommendation that 0x be accepted as felt252 and that I should add a 0 to 0x to be accepted

metalboyrick: yep, aligned on that end

metalboyrick: please make sure `0x0x0` does not work

KevinMB0220: Ok I'll check that it aligns with that

metalboyrick (changes requested):
seems invalid address still accepted, should not be accepted

expected behavior:
- `0x` treated as `0x0`
- `0x0x0` should be error
- `x0` should be error
- `0x<something>` should be correct

image

KevinMB0220:
> seems invalid address still accepted, should not be accepted
> 
> expected behavior:
> 
> * `0x` treated as `0x0`
> * `0x0x0` should be error
> * `x0` should be error
> * `0x<something>` should be correct
> 
> image

The validation is now ready so that what you requested is not allowed and only valid addresses are allowed.

Nadai2010 (changes requested):
Hi @KevinMB0220 this is still malfunctioning, can check please? [It doesn't do this]( https://github.com/Scaffold-Stark/scaffold-stark-2/pull/398#pullrequestreview-2519564576) for me.

Captura desde 2024-12-23 07-27-06
Captura desde 2024-12-23 07-26-22
Captura desde 2024-12-23 07-26-43
Captura desde 2024-12-23 07-26-55

KevinMB0220:
> Hi @KevinMB0220 this is still malfunctioning, can check please? [It doesn't do this](https://github.com/Scaffold-Stark/scaffold-stark-2/pull/398#pullrequestreview-2519564576) for me.
> 
> Captura desde 2024-12-23 07-27-06 Captura desde 2024-12-23 07-26-22 Captura desde 2024-12-23 07-26-43 Captura desde 2024-12-23 07-26-55

Ok i check

packages/nextjs/components/scaffold-stark/Address.tsx
```diff
@@ -60,6 +65,19 @@ export const Address = ({
...
   );
 
+  const isValidHexAddress = (value: string): boolean => {
+    if (value.toLowerCase() === "0x") {
+      value = "0x0";
+    }
+
+    if (value.toLowerCase() === "0x0x0") {
+      return false;
+    }
```
metalboyrick, line 75: `0x0x0` is just an example, what if the address is like `0x0x141240x` ? this is supposed to be invalid. Have to account for all the cases as well.

metalboyrick (changes requested):
seems like this is not the spot to fix. might want to take a look at the `InputBase` and `Input` component to solve this issue.

metalboyrick:
@KevinMB0220 what we want to do is to check for the input, not the address display

KevinMB0220:
> @KevinMB0220 what we want to do is to check for the input, not the address display

Can I change something in the faucet since the input address is declared there, so the validation is more direct and functional there?

metalboyrick:
> > @KevinMB0220 what we want to do is to check for the input, not the address display
> 
> Can I change something in the faucet since the input address is declared there, so the validation is more direct and functional there?

yep sure!

packages/nextjs/components/scaffold-stark/Input/AddressInput.tsx
```diff
@@ -21,7 +21,25 @@ export const AddressInput = ({
...
+      const sanitizedValue = newValue.toLowerCase();
+
+      if (sanitizedValue === "0x") {
+        onChange("0x0" as Address);
+        return;
+      }
+
+      if (/^0x.*0x/.test(sanitizedValue.slice(2))) {
+        return;
+      }
```
metalboyrick, line 33: this still does not cover the cases. would be best to just use a regex maybe? e.g. invalid stuff like `0x1235x21412xyz` might happen.

metalboyrick (changes requested):
sorry for the long winded review but we need to get this right. Asides from the `AddressInput` component you would also need to validate the one on the Debug Page (which doesn't use this component)

metalboyrick:
> sorry for the long winded review but we need to get this right. Asides from the `AddressInput` component you would also need to validate the one on the Debug Page (which doesn't use this component)

i think you can fix the `AddressInput` component first, let's think about how to abstract the validations to the inputs on the debug page later on (cc @Nadai2010 )

packages/nextjs/components/scaffold-stark/Input/AddressInput.tsx
KevinMB0220: Ok i working in this

metalboyrick (approved):
looks good to me!

Nadai2010 (approved):

PR #404, opened 2024-12-22
Fixed [BUG] u256> display for non-zero u256

### LINKED ISSUE #344: [BUG] `u256>` display for non-zero `u256`
## Issue Overview

`u256>` display for non-zero `u256`, This will happen to all non-zero type-wrappers. We expect only `u256 will appear on these cases`

image

- Parse generic type and extract the insides.

jedstroke:
# [BUG] u256> display for non-zero u256

WhatsApp Image 2024-12-22 at 8 55 12 PM

Fixes #344 

## Types of change

- [ ] Feature
- [x] Bug
- [ ] Enhancement


jedstroke:
@Nadai2010 @metalboyrick

packages/nextjs/components/scaffold-stark/Input/EtherInput.tsx
```diff
@@ -129,9 +129,9 @@ export const EtherInput = ({
 
   return (
     <InputBase
-      name={name}
+      name={name?.replace(">", "")}
```
metalboyrick, line 132: can't fix it like this because this will break happy paths

metalboyrick (changes requested):

metalboyrick:
GM @jedstroke  seems like we will not be able to finish the PR in due time. Therefore we would not be able to consider this PR as an ODHack 11 PR, but you can still work on it but the rewards will not be part of the ODHack11 reward. 

Really appreciate the attempt!

Nadai2010:
Hello @jedstroke thank you for your contribution, this PR needs a lot of changes that explain many files uploaded that should not be, we close the issue by this ODHack

PR #408, opened 2025-01-02
format event data when needed

iossocket:
# Format event data

## Types of change

- [ ] Feature
- [ ] Bug
- [x] Enhancement

## Comments
Almost all the event parameters have been parsed correctly. provide an option to format both the address type and the nested address type.

### DIFF
--- a/packages/nextjs/hooks/scaffold-stark/useScaffoldEventHistory.ts
+++ b/packages/nextjs/hooks/scaffold-stark/useScaffoldEventHistory.ts
@@ -205,13 +205,10 @@ export const useScaffoldEventHistory = <
         );
         const args = parsed.length ? parsed[0][eventName] : {};
         const { event: rawEvent, ...rest } = event;
-        const parsedArgs = format
-          ? parseEventData(args, rawEvent.members)
-          : null;
         return {
           type: rawEvent.members,
           args,
-          parsedArgs,
+          parsedArgs: format ? parseEventData(args, rawEvent.members) : null,
           ...rest,
         };
       });
--- a/packages/nextjs/hooks/scaffold-stark/useScaffoldEventHistory.ts
+++ b/packages/nextjs/hooks/scaffold-stark/useScaffoldEventHistory.ts
@@ -18,6 +18,7 @@ import { devnet } from "@starknet-react/chains";
 import { useProvider } from "@starknet-react/core";
 import { hash, RpcProvider } from "starknet";
 import { events as starknetEvents, CallData } from "starknet";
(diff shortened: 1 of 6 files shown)


metalboyrick (commented):
are you going to fix the filter bug in this PR?

iossocket:
> are you going to fix the filter bug in this PR?

That will be in another PR

packages/nextjs/hooks/scaffold-stark/useScaffoldEventHistory.ts
```diff
@@ -201,14 +204,20 @@ export const useScaffoldEventHistory = <
...
+        const { event: rawEvent, ...rest } = event;
+        const parsedArgs = format
+          ? parseEventData(args, rawEvent.members)
+          : null;
         return {
+          type: rawEvent.members,
           args,
-          ...event,
+          parsedArgs,
+          ...rest,
```
metalboyrick, line 215: nit: can simplify the selected lines to the following ?

```tsx
 return {
          type: rawEvent.members,
          args,
          parsedArgs: format && parseEventData(args, rawEvent.members),
          ...rest,
   }
```

metalboyrick (changes requested):
Seems there's a small piece that kinda breaks the hook. I got the following error when trying to fetch events that are NOT EMPTY:

image

Steps to reproduce:
- Setup and deploy contracts as-is out of the box with devnet.
- Implemented the following in main page:
```ts
  const { data: events } = useScaffoldEventHistory({
    contractName: "Strk",
    eventName: "openzeppelin::token::erc20_v070::erc20::ERC20::Transfer",
    format: false,
    fromBlock: 0n,
  });
```
- Initially it works if the returned events is empty.
- Performed a Transfer with STRK token.
- The above error happens.

NOTE: when `format` is false or true the above error happens anyways. I think it's an unexpected `undefined` in one of the event object key-value pairs, might need to do some digging.

iossocket:
@metalboyrick thanks for pointing out this issue, it is caused by wrong abi type defined, and fixed in this commit: https://github.com/Scaffold-Stark/scaffold-stark-2/pull/408/commits/f75a263bbfa054d4207f9f89e60010e474c8e9bd

packages/nextjs/hooks/scaffold-stark/useScaffoldEventHistory.ts
iossocket: updated

metalboyrick (changes requested):
Still there ser, did the same STR and got the same error

metalboyrick (approved):
my bad seems i forgot to pull or sth, works fine, lgtm!

image

PR #413, opened 2025-01-07
Yarn declared

### LINKED ISSUE #409: Add new command: yarn declare --network sepolia for contract deployment
## Issue Overview
Need to implement a new command `yarn declare --network sepolia` to support contract declaration on Sepolia testnet.
- Add new script in package.json for `yarn declare`
- Support network flag `--network` with `sepolia` option

Buuh2511:
# yarn declared & export mainetFork as const rather than chains.Chain

Fixes #409 

## Types of change

- [x] Feature
- [ ] Bug
- [ ] Enhancement


packages/snfoundry/scripts-ts/declare-contract.ts
jrcarlos2000, line 1: steps for deployment :
 1. declare if not declared
 2. deploy

we should be able to reuse the same functions for `yarn declare --network --fee` and `yarn deploy --network --fee` 

Please find the intersection so to reuse this functions on the `deploy-contract` script

packages/nextjs/supportedChains.ts
```diff
@@ -2,7 +2,7 @@ import * as chains from "@starknet-react/chains";
 
 // devnet with mainnet network ID
 const mainnetFork = {
-  id: BigInt("0x534e5f4d41494e"),
+  id: BigInt("393402133025997798000961"), //devnet id
```
0xquantum3labs, line 5: a fork holds the ID of the

0xquantum3labs, line 5: if this is the same id as devnet, what is the difference between this `mainnetFork` and `chains.devnet` ?

0xquantum3labs: the whole thing would be the same

packages/snfoundry/scripts-ts/deploy-contract.ts
```diff
@@ -89,6 +111,37 @@ const declareIfNot_NotWait = async (
   };
 };
 
+const declareContract = async (): Promise<void> => {
+  try {
+    const contract = "YourContract";
```
0xquantum3labs, line 116: this shouldnt be hardcoded here

packages/snfoundry/scripts-ts/deploy-contract.ts
```diff
@@ -89,6 +111,37 @@ const declareIfNot_NotWait = async (
   };
 };
 
+const declareContract = async (): Promise<void> => {
```
0xquantum3labs, line 114: i believe this must receive a contract name

0xquantum3labs (changes requested):

0xquantum3labs:
i believe we should instead add a function `declare` that we can use the same way we use the `deployContract()` function on the [this file ](https://github.com/Scaffold-Stark/scaffold-stark-2/blob/main/packages/snfoundry/scripts-ts/deploy.ts)


and keep using the `yarn deploy` command. 

note : do not overwrite the deployments array [here](https://github.com/Scaffold-Stark/scaffold-stark-2/blob/127d742e61c4d3abbb875f3d3280529e47422c7e/packages/snfoundry/scripts-ts/deploy-contract.ts#L55) 
can consider creating another array called `declarations` , we dont export it as the `deployments` into the nextjs package

please note that when declaring a contract, we dont need a set of constructor calldata unlike when deploying.

jrcarlos2000 (changes requested):

0xquantum3labs:
closing this for now, is a minor feature that we can achieve on a mid-step on `deployContract` , it will add some complexity to the monorepo deployment logic as this wouldnt update the `deployedContracts.ts` file

PR #419, opened 2025-01-17
fix: yarn deploy network args

metalboyrick:
# Task name here

- Errors out invalid arguments
- Fixed inverted `yarn deploy:no-reset`
- Fixed `no-reset` not having any effects

## Types of change

- [ ] Feature
- [] Bug
- [ ] Enhancement


packages/snfoundry/scripts-ts/helpers/deploy-wrapper.ts
```diff
@@ -10,33 +10,40 @@ interface CommandLineOptions {
...
 }
 
-const argv = yargs(process.argv.slice(2))
-  .options({
-    network: { type: "string" },
-    fee: { type: "string", choices: ["eth", "strk"], default: "eth" },
-    reset: {
+function main() {
+  const argv = yargs(process.argv.slice(2))
+    .option("network", { type: "string", default: "devnet" })
```
Nadai2010, line 15: I think we could leave a `choice` here to avoid the failure in an incorrect network by the user with a correct understanding of the error in the syntax, as is in `fee`.

`choices: ["sepolia", "mainnet", "devnet"]`

imagen

Nadai2010 (changes requested):
Thank you for resolved the isuue @metalboyrick, The `reset` and `no-reset` logic works perfectly.

packages/snfoundry/scripts-ts/helpers/deploy-wrapper.ts
metalboyrick: updated, thanks for the input!

Nadai2010 (approved):

PR #421, opened 2025-01-19
fix: correct address regex in AddressInput

### LINKED ISSUE #420: [BUG] Starknet addresses can't be pasted into faucet input
# Issue Overview
Typical Starknet address cannot be pasted into the faucet because the input has a validator that checks if address length equals to 42.
It's not correct because in Starknet addresses are felt and can be 32 bytes long (which corresponds to address length of 66 including '0x' prefix).
Also I believe it's not correct to check for minimal length in Starknet because leading zeroes are often dropped and 0x1 is a technically correct address.
The bug was introduced in #398 
Relevant lines of code: https://github.com/Scaffold-Stark/scaffold-stark-2/blob/6e9eb830bdb8744348ee6f26c382f85c1a2d916c/packages/nextjs/components/scaffold-stark/Input/AddressInput.tsx#L31-L40
# Proposed Solutions or Ideas
I guess it will be sufficient to test for `/^0x[0-9A-Fa-f]{1,64}$/` and remove all redundant checks like `/0x.*0x/` and `sanitizedValue.length !== 42`

kfastov:
# Fix address check in AddressInput

Fixes #420

## Types of change

- [ ] Feature
- [x] Bug
- [ ] Enhancement

## Comments (optional)
Changes made:
1. Updated the regex to `/^0x[a-f0-9]{1,64}$/` which:
   - Matches `0x` prefix
   - Followed by 1-64 hex characters (a-f, 0-9)
   - No need for capture groups or additional checks
2. Removed the redundant length check since it's handled by the regex
3. Removed the redundant double-0x check since the regex ensures only one 0x at the start
4. Fixed formatting issues in `deploy-wrapper.ts`

The validation is now more appropriate for Starknet addresses while being simpler and more maintainable.

### DIFF
--- a/packages/nextjs/components/scaffold-stark/Input/AddressInput.tsx
+++ b/packages/nextjs/components/scaffold-stark/Input/AddressInput.tsx
@@ -28,14 +28,8 @@ export const AddressInput = ({
         return;
       }
 
-      const isValid =
-        /^(0x)([a-fA-F0-9]{40})$/.test(sanitizedValue) &&
-        !/0x.*0x/.test(sanitizedValue);
+      const isValid = /^0x[a-f0-9]{1,64}$/.test(sanitizedValue);
       if (!isValid) {
-        return;
-      }
-
-      if (sanitizedValue.length !== 42) {
         return;
       }
 
--- a/packages/snfoundry/scripts-ts/helpers/deploy-wrapper.ts
+++ b/packages/snfoundry/scripts-ts/helpers/deploy-wrapper.ts
@@ -28,7 +28,7 @@ function main() {
 
   if (argv._.length > 0) {
     console.error(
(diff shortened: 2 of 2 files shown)


Nadai2010 (changes requested):
Hi thanks for the input @kfastov, for the action flow to be correct ‘in green’ and formatted properly, you should check the `yarn format:check` and `yarn format`.

The rest is fine by me, @metalboyrick check it here please

kfastov:
@Nadai2010 The formatting issue is in the mainstream, but no problem, I'll fix it

Nadai2010:
Ah I just uploaded that correct format from before, it looks like you just took that version, I'll close my PR and leave it in this one thanks

Nadai2010 (approved):

0xquantum3labs (approved):

0xquantum3labs:
thanks for your contribution @kfastov 🤝

PR #427, opened 2025-01-26
Keplr connector

iossocket:
# Keplr connector

## Types of change

- [x] Feature
- [ ] Bug
- [ ] Enhancement

image

### How to test on devnet
1. start the fork chain: `yarn chain --fork-network https://starknet-mainnet.public.blastapi.io/rpc/v0_7`
2. change main net rpc url to `http://127.0.0.1:5050/rpc`. `Settings` -> `Advanced` -> `Change Endpoints`
image


Nadai2010 (approved):
Hi @iossocket , thank you very much for your contribution.

I'm checking on Testnet and it works perfectly, in keplr I see that it doesn't work from devnet, but we can use isFork, so I think it's fine, since if we remove it in devnet mode we wouldn't be able to play with this feature.

imagen

PR #433, opened 2025-01-31
Cache getClassHash request and result

iossocket:
# Cache getClassHash request and result

## Types of change

- [ ] Feature
- [ ] Bug
- [x] Enhancement

## Comments

We use `useDeployedContractInfo` in many places, then the `getClassHashAt` API is triggered significantly more times than actually needed. When we provide provider url in env file, it will cause http 429 error.
image
image

In this PR, added a singleton instance to handle the cache,  it will reuse the api call as well.


Nadai2010 (approved):

Nadai2010:
Good job for me @iossocket thank you

PR #444, opened 2025-02-20
Enhance developer annotations.

ccolorado:
# Task name here
Enhance developer annotations.

Fixes #relevant-issue-here

## Types of change

- [ x] Enhancement

## Comments (optional)
1. While the initial intention was to add a help command, in practice it collides with the `yarn help` command. 
Instead added the command `usage` to present the information. At this point it might be better to just add a table to the README.md as `usage` is not a standard. 
 
2.  Added a `.env.example` file as it seems to be missing.


.env.example
metalboyrick, line 1: thanks for the PR, the env is located at each submodule, thus they are there. you can remove the root one

README.md
metalboyrick, line 1: might want to indicate how to run the `usage` command itself

metalboyrick (changes requested):
thanks for the PR, might need some changes, feel free to discuss as well

README.md
ccolorado: I think the readme file might be the only appropriate place suitable for this list. Even trying to add a `--help` script would be ambiguous and confusing. I could turn it into a markdown table. 

Perhaps the HELP.md file could be shown at the end on the `npx create-stark` ?

.env.example
ccolorado: I noticed the  *_MAINNET variables are missing from the foundry examples. I can see omitting them can avoid accidents.

This looks  to me like a very probable  place for devs to look at while trying to figure  out how to do deployments, perhaps a pointing to the documentation here can prevent devs from getting lost. 

# For the full deployment configuration options visit:  https://docs.scaffoldstark.com/deploying/deploy-smart-contracts

README.md
metalboyrick: yep we can just leave it on the readme

The thing is some of the audience are people who are very new to coding in general, thus it would be more helpful if we sort of baby step it.

.env.example
metalboyrick: yes @ccolorado that would be good

HELP.txt
metalboyrick, line 1: let's just remove this and put in readme

metalboyrick:
left replies, sorry for the delay @ccolorado

HELP.txt
ccolorado: Removed file and added as a markdown table.

.env.example
ccolorado: Added annotation, Please mark as resolved if no further comments.

ccolorado:
Los cambios solicitados ya fueron aplicados, pero ahora veo un conflicto con main. 
Crees que mi nueva sección deba de ir antes o después de la de `## Network Configuration Centralization`

metalboyrick:
@ccolorado anywhere is fine, does not matter in this case

(sorry I do not speak Spanish, used a translator to read your message)

metalboyrick:
also apologies for the late reply

ccolorado:
Conflict is resolved.

gianalarcon (approved):

Issue #442, opened 2025-02-20
[BUG] Make sure that framework is compatible with Starknet Devnet 0.2.4


metalboyrick:
Starknet Devnet 0.2.4 uses transaction V3 and Scaffold-Stark currently has only partial support for it, we would like to extend it to all of our contract calls.

Make sure the following warning(s) are gone!

![Image](https://github.com/user-attachments/assets/b7ee62f0-23c9-44d6-b84c-82c7dadf8f46)

Scope: 
- Bump to starknet 0.2.4
- Bump starknet.js to latest compatible version (package.json)
- Refactor contract interaction code to fully support Transaction V3 format
- Ensure all deprecated method calls are replaced with V3-compatible alternatives
- Add/update tests to verify proper Transaction V3 implementation

Joewizy:
Can I work on this, please?

KevinMB0220:
Hello, my name is Kevin, I am part of Dojo Coding, I have experience as a full stack developer, I have contributed to various oppen source projects, I would follow these steps to resolve the issue:

To refactor the .then() syntax into async/await within deploy.ts, I will:
Refactor deploy.ts:
Replace all .then() chains with async/await for better readability.
Handle errors using try-catch blocks for cleaner error management.
Check Other Files:
Search for similar .then() patterns across the project.
Apply consistent async/await syntax throughout.

Testing:
Ensure functionality remains unchanged.
Test deployment flow for any regressions.
This improves maintainability and code readability. 🚀

Nadai2010:
Hi @manlikeHB , this issue has been assigned to you, we are looking forward to your job, thanks

Nadai2010:
Hello @manlikeHB any news about the issue?

Dalongcoder0522:
@Nadai2010 please assign this issue to me.

PR #449, opened 2025-03-01
snfoundry: Update dependencies

gianalarcon:
# Task name here

Fixes #relevant-issue-here

- Updated GitHub workflow configuration to automatically adapt to new dependency versions.
- Successfully performed comprehensive script testing and manual verification
image
- Maintained/kept starknet-devnet 2.3.0 version to prevent warning messages

## Types of change

- [ ] Feature
- [ ] Bug
- [x] Enhancement

## Comments (optional)
- This PR serves as prerequisite work for issue #442
- Before merging, the docker container needs to be updated with the latest dependency versions to ensure full compatibility.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2457,14 +2457,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@scure/base@npm:~1.1.3":
-  version: 1.1.7
-  resolution: "@scure/base@npm:1.1.7"
-  checksum: d9084be9a2f27971df1684af9e40bb750e86f549345e1bb3227fb61673c0c83569c92c1cb0a4ddccb32650b39d3cd3c145603b926ba751c9bc60c27317549b20
-  languageName: node
-  linkType: hard
-
-"@scure/base@npm:~1.1.7, @scure/base@npm:~1.1.8":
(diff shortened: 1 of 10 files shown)


yarn.lock
Nadai2010, line 1: I think this file should not be changed, since no dependency has been updated that influences it and I see that the dependency of "@starknet-react/core@npm:3.7.2", chain...

Nadai2010 (changes requested):
I like the change to the action. These updates improve the workflow by using GitHub Actions for installing tools (`setup-scarb` and `setup-snfoundry`) instead of custom scripts.

I've also tested with devnet version 0.2.4/0.2.3, scarb 2.10.1, and snfoundry 0.38.0, in preparation for the mainnet release 0.14.0 on March 17th. Currently, devnet is not ready, as it is still in the following PR:

[PR for update 0.13.4 devnetrs](https://github.com/0xSpaceShard/starknet-devnet/pull/697)

However, it is working fine on Sepolia:

[Sepolia Contract Link](https://sepolia.starkscan.co/contract/0x0423ba09341dc0a0cb112e1b776cf74ed52359d394c1944bb51af042f44bf11a)

imagen

So, the proposed change is the best option right now. The README should also be updated to reflect the new versions.

Once the ODBuild reviews are finished, it can be merged.

Nadai2010 (approved):

PR #456, opened 2025-03-17
[BUG] Add pwa in nextjs dependencies

Nadai2010:
# Task name here

Integrate PWA into NextJS dependencies, otherwise the "yarn deploy:yolo" options will fail.

imagen
Captura desde 2025-03-17 09-55-41

## Types of change

- [x] Bug


metalboyrick (approved):
So far looks good, 

If we want to continue making it a full installable PWA, we can do that: https://ducanh-next-pwa.vercel.app/docs/next-pwa/getting-started#step-2-add-a-web-application-manifest

PR #458, opened 2025-03-21
build: update dependencies to latest version

metalboyrick:
# Task name here

- Bump Dependencies on the following
  - Starknet React => update to 3.7.2
  - StarknetJS => 6.23.1
  - Next JS => 15
  - React => 19
  
- Move `next-pwa` into `nextjs` package

NOTE: you need to reinstall `node_modules` with the following commands if you encounter SWC errors:
```bash
yarn install --force
```


packages/nextjs/package.json
```diff
@@ -24,8 +24,8 @@
     "@keplr-wallet/unit": "^0.12.175",
     "@radix-ui/react-icons": "1.3.0",
     "@radix-ui/themes": "2.0.3",
-    "@scaffold-stark/stark-burner": "0.1.9",
```
gianalarcon, line 27: we should use the latest version of burner in order to support v3 tx

metalboyrick: fixed good ser

PR #465, opened 2025-03-28
feat:fix the yarn deploy error

Dalongcoder0522:
# Task name here
feat:fix the yarn deploy error
Fixes #relevant-issue-here

## Types of change

- [ ] Feature
- [X] Bug
- [ ] Enhancement

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
 scarb 2.11.3
 starknet-foundry 0.40.0
-starknet-devnet 0.2.4
+starknet-devnet 0.3.0
--- a/packages/snfoundry/package.json
+++ b/packages/snfoundry/package.json
@@ -29,7 +29,7 @@
     "dotenv": "^16.3.1",
     "envfile": "^6.18.0",
     "prettier": "^2.8.8",
-    "starknet": "^6.24.0",
+    "starknet": "^7.0.0-beta.4",
     "toml": "^3.0.0",
     "tslib": "^2.6.2"
   },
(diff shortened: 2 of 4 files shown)


Nadai2010 (changes requested):
Hello, thanks for the work! Here are some issues:

## Devnet  
- There's a small syntax error in `deploy-contract.ts`, with an `"unknown"` type. It can be quickly fixed to allow deployment.  
- (Note: The first declaration + deploy cost 33 STRK, while subsequent deployments cost 1 STRK.)  

Captura desde 2025-03-31 08-09-31

---

### Invoke 
- The DebugUI has some display errors in `displayVariable.tsx`.  
- When trying to execute an `invoke`, we see the `tx_v3` error.  
- It seems that gas values for L2 and L1 are not being set correctly, leading to an error: `"l1_data_gas": undefined`.  
- (Even though there have been issues on Sepolia regarding fee calculations for Blobs—see the open issue [here](https://github.com/Blobscan/blobscan/issues/748)—I don't think it affects this specific case.)  

Captura desde 2025-03-31 08-09-59

Captura desde 2025-03-31 08-10-21

Captura desde 2025-03-31 08-10-30

---

## Sepolia
- Deployment does not work with **Braavos** (it works fine with Argent).  
- We see `invalid_tx` failures and account validation errors (tested with two different accounts).  
- Once deployed, invoking works fine with both Braavos and Argent—it's only the deployment that fails.  

Captura desde 2025-03-31 08-11-33

- Update `Starknet.js` to the more stable [V7.0.1](https://github.com/starknet-io/starknet.js/releases/tag/v7.0.1) instead of the beta version.

PR #472, opened 2025-04-01
Gian/update - fix:nextjs error and predeployed ETH STRK name() symbol() not working workaround

Dalongcoder0522:
# Task name here
fix:nextjs error and predeployed ETH STRK name() symbol() not working workaround
Fixes #relevant-issue-here

## Types of change

- [ ] Feature
- [X] Bug
- [ ] Enhancement

## Comments (optional)
The Bravvos and Argent wallet still not working when deploying contract to sepolia

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
 scarb 2.11.3
 starknet-foundry 0.40.0
-starknet-devnet 0.2.4
+starknet-devnet 0.3.0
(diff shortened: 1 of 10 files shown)


gianalarcon:
Few Comments

- **On devnet:** Still getting the issue of declare + deploy, how could it cost more than 30 STRKs
image

- **On Sepolia:** not working as you mentioned

PR #492, opened 2025-04-17
refactor: add warn for empty RPC Provider URL config in next .env

tu11aa:
# Add warn for empty RPC Provider URL config in next .env

## Types of change

- [ ] Feature
- [ ] Bug
- [X] Enhancement

## Changes
- Add warn for empty RPC Provider URL config in next .env


gianalarcon (approved):
The issue here why the provider is showing `rate limits`? 
So there is a bug in the toolkit that makes a lot of calls/ 
Scaffold-stark suppose to do few calls/invokes, we shouldn't get this kind of error

PR #497, opened 2025-04-18
chore: remove console log error when api fail

### LINKED ISSUE #487: Bug: Starknet.id feature emits error on client side
## Description
After a fresh deployment, Debug contract UI gets erros because of Starknet.id feature

Image

## Solution
Fix files related to this issue

BaoHG1508:
# Task name here

Fixes #487

## Types of change

- [x] Bug

## Comments (optional)

- When the Starknet API cannot find data for a given address, it returns a 400 Bad Request with a message like “No data found for the given address.” This is expected behavior and not necessarily an error.

- In this PR, I’ve removed the explicit error handling that previously resulted in noisy logs. Please note that browser developer tools (e.g. Chrome DevTools) may still display a network error for the failed request — this is outside of our control.


gianalarcon:
This PR should point to `develop` branch, not to `main` branch, thanks

PR #501, opened 2025-04-21
Refactor: refactor copying flow from Scaffold-stark to Create-stark

tu11aa:
# Refactor: refactor copying flow from Scaffold-stark to Create-stark

Fixes #relevant-issue-here

## Types of change

- [ ] Feature
- [X] Bug
- [X] Enhancement

## Changes
- d89f4098b0d3841eed34f7b356251afe8bb84239: exclude unnecessary files and dirs


gianalarcon:
- Why not direct edit on `release-create-stark.yml` file? Can edit direct there.
- We need both repositories(scaffold-stark and create-stark) to maintain the same version, then the workflow should ensure that when one repo's version is bumped, the same exact version bump is applied to the other repo.

gianalarcon:
Probably also need to fix this PR 
https://github.com/Scaffold-Stark/create-stark/pull/16/files

PR #523, opened 2025-05-01
docs: update readme to use asdf install

metalboyrick:
Recently I have discovered that simply running `asdf install` will give you all the needed dependencies. I believe this could help with the dependency installation. Please verify, thanks !


gianalarcon:
Looks good, it can be used after `starkup`. By default starkup install the latest stable version of `asdf`, `scarb` and `snfoundry`.
We can use `asdf install` to make sure we install the specfic required version of .tool-versions

Suggest to check full readme and in order to explain and show the way to use `starkup` and `asdf install` together.

metalboyrick:
pushed an update @gianalarcon see if that wording sounds ok. I just say that this can be used if people experience version errors with `starkup`.

gianalarcon (approved):

PR #546, opened 2025-05-30
Re-Connect wallet when refetch

### LINKED ISSUE #539: SS2 : [Bug] Wallet got disconnected after a F5/refresh of UI
## Issue Overview

After you connect your wallet and refresh the UI view, it disconnects your wallet and we need to connect again in order to interact with the dapp

Keep a longer connection of the wallet, disconnect after 24 hours of inactivity o suggest a better way to manage it.

Buuh2511:
# Re-Connect wallet when refetch
Fixes #539 

## Types of change

- [ ] Feature
- [x] Bug
- [x] Enhancement

## Comments (optional)
- Do event when disconnect wallet => listen event to tracking if any connection
- Update useAtuconnect
- Update test for useAutoconnect

### DIFF
--- a/packages/nextjs/hooks/scaffold-stark/__tests__/useAutoConnect.test.ts
+++ b/packages/nextjs/hooks/scaffold-stark/__tests__/useAutoConnect.test.ts
@@ -1,115 +1,79 @@
-import { renderHook } from "@testing-library/react";
+import { renderHook, act, waitFor } from "@testing-library/react";
 import { useAutoConnect } from "../useAutoConnect";
-import { useConnect } from "@starknet-react/core";
 import { useReadLocalStorage } from "usehooks-ts";
 import scaffoldConfig from "~~/scaffold.config";
 import { burnerAccounts } from "@scaffold-stark/stark-burner";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
-// Mock the dependencies
+const mockConnect = vi.fn();
+
+const burnerConnector = {
+  id: "burner-wallet",
+  ix: 1,
+  ready: true,
+  burnerAccount: burnerAccounts[1],
+};
+
+vi.mock("@starknet-react/core", async (importOriginal) => {
+  const actual = (await importOriginal()) as Record<string, any>;
+  return {
+    ...actual,
+    useConnect: () => ({
+      connect: mockConnect,
+      connectors: [burnerConnector],
+    }),
+    useAccount: () => ({
+      account: null,
+      address: undefined,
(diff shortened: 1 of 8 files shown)


packages/nextjs/hooks/scaffold-stark/__tests__/useAutoConnect.test.ts
```diff
@@ -1,115 +1,79 @@
...
   });
 
-  it("should auto-connect if walletAutoConnect is enabled and a saved connector exists", () => {
-    vi.mocked(useReadLocalStorage).mockReturnValue({ id: "wallet-1" });
-
-    renderHook(() => useAutoConnect());
-
-    expect(mockConnect).toHaveBeenCalledWith({
-      connector: expect.objectContaining({ id: "wallet-1" }),
+  it("should auto-connect to the burner wallet if savedConnector exists", async () => {
```
gianalarcon, line 55: Instead of delete, lets update the other tests like
- it("should auto-connect if walletAutoConnect is enabled and a saved connector exists")
- it("should not auto-connect if walletAutoConnect is disabled"
- it("should not connect if there is no saved connector"
- it("should not connect if saved connector is not found in the connectors list"

packages/nextjs/scaffold.config.ts
```diff
@@ -13,7 +13,7 @@ export type ScaffoldConfig = {
 };
 
 const scaffoldConfig = {
-  targetNetworks: [chains.devnet],
+  targetNetworks: [chains.sepolia],
```
gianalarcon, line 16: lets keep on devnet

gianalarcon (changes requested):

packages/nextjs/hooks/scaffold-stark/__tests__/useAutoConnect.test.ts
Buuh2511: Hello sir.
I just want you to confirm a little bit. I changed the implementation of the autoconnect function so the original test cases will be wrong and I updated them to new test cases. You still want to keep the original test cases and add new test cases even if they are describing wrong, right?

packages/nextjs/scaffold.config.ts
Buuh2511: has returned to devnet

gianalarcon (approved):

metalboyrick (changes requested):
gm ser Buu! thanks for the PR! Minor comments from my side

metalboyrick (changes requested):

PR #549, opened 2025-05-31
chore: fixed bug for untyped enum

### LINKED ISSUE #538: [Bug] Improve handling of untyped enums
## Issue Overview

On this [branch](https://github.com/Scaffold-Stark/scaffold-stark-2/tree/untyped-enum)
We introduced a untyped enum

```cairo
/// Represents the possible states of a Multisig transaction.
#[derive(Copy, Drop, Serde, starknet::Store, PartialEq, Debug)]
pub enum TransactionState {
    #[default]
    NotFound,
    Pending,
    Confirmed,
    Executed,
}
```

When writing with enums which are not typed, `debug-ui` will require us to choose a variant/option and input something into the input-box. Check below

Image

The ABI genereted for this enum looks something like this :
https://github.com/Scaffold-Stark/scaffold-stark-2/blob/d71930f7556a7f6f1bdea0e13c301a4e6966d7e7/packages/nextjs/contracts/deployedContracts.ts#L78-L99

The `Debud contracts` should handle this kind of enums properly, it should work without having to input anything on the input-box. As long a we choose a variant/option it should be ok to `send` 

In case it actually needs an input, like `()` input, it should be handled internally when creating the Enum on the react hook.

Imdavyking:
# Improve handling of untyped enums

Fixes #538

## Types of change

- [ ] Feature
- [x] Bug
- [ ] Enhancement

Closes #538

### DIFF
--- a/packages/nextjs/app/debug/_components/contract/ContractInput.tsx
+++ b/packages/nextjs/app/debug/_components/contract/ContractInput.tsx
@@ -114,7 +114,7 @@ export const ContractInput = ({
           // @ts-ignore
           abiMember={abi?.find(
             // @ts-ignore
-            (member) => member.name === paramType.type
+            (member) => member.name === paramType.type,
           )}
           isDisabled={isDisabled}
         />
--- a/packages/nextjs/app/debug/_components/contract/ContractInput.tsx
+++ b/packages/nextjs/app/debug/_components/contract/ContractInput.tsx
@@ -113,7 +113,7 @@ export const ContractInput = ({
           // @ts-ignore
           abiMember={abi?.find(
             // @ts-ignore
-            (member) => member.name === paramType.type
+            (member) => member.name === paramType.type,
           )}
           isDisabled={isDisabled}
         />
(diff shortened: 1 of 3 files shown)


gianalarcon:
Looks good to me. Please apply `yarn format`

Imdavyking:
thank you
ran yarn format

packages/nextjs/app/debug/_components/contract/ContractInput.tsx
```diff
@@ -98,6 +98,9 @@ export const ContractInput = ({
       !isCairoResult(paramType.type) &&
       !isCairoOption(paramType.type)
     ) {
+      if (paramType.type == "()") {
```
metalboyrick, line 101: can we abstract this check into an utility function? we can call it `isCairoVoid`. so that we have convention consistency, similar to line 99.

metalboyrick (changes requested):
Thanks for the PR! Functionality works good!

Some minor comments, and also do not forget to update the test cases if there are any!

Imdavyking:
Done that and wrote tests

gianalarcon (approved):

gianalarcon:
Thank you so much for your contribution and joining ODHack 14 @Imdavyking

PR #552, opened 2025-06-02
Add get_var_non_zero and set_var_non_zero for u256 support in YourCon…

Haroldwonder:
# Get and Set not working for non-zero u256 data type

Fixes ODHack14

Added `get_var_non_zero` and `set_var_non_zero` functions to `YourContract.cairo` to properly handle the `u256` data type by splitting it into two `u128` parts (low and high). Updated the contract interface, storage, and implementation to support this.


gianalarcon:
Thanks for contributing
This PR doesnt solve the issue
We expect to parse the Non-zero 256 value using typescript so we can interact with the contract using `Debug Contracts`, The issue description showed a screenshot that have to be fixed

PR #564, opened 2025-06-13
Fix: Shortened Event Argument Names in useScaffoldEventHistory Hook

### LINKED ISSUE #543: SS2: Shorten Event Argument Names for `useScaffoldEventHistory`
## Overview

Currently, to be able to listen to events using the `useScaffoldEventHistory` hook, you need to input a long string for the event name to do so, else the linter will mark it as a TypeScript Error

Image:

Image

This is also not very good for the autocomplete, as we cannot exactly see what the event name is

Image:

Image

The end goal is to just use the event name to define the event name within this hooks, e.g.

```tsx
const data = useScaffoldEventHistory({
    contractName: "Strk",
    eventName: "Transfer",
    // ...
})
```

## Ideas

Take a look at the abi parser and find out the issue from there.

OWK50GA:
# SS2: Shorten Event Argument Names for useScaffoldEventHistory

Fixes #543 

## Types of change

- [ ] Feature
- [ ] Bug
- [x] Enhancement

## Comments 

I constructed a new type to interact with another type, and extract the base name of the argument type it is given, so that 'Transfer' can be extracted from the long event name for instance, and the auto-complete still recognizes it. I changed this type in the UseScaffoldEventHistoryConfig (more like added, with an or separator so that both methods can be used). Then in the readEvents function, I made sure to check either the basename or the full name of the events for matching with the eventName argument.

Details in code, it was a very small change

### DIFF
--- a/packages/nextjs/hooks/scaffold-stark/__tests__/useScaffoldEventHistory.test.ts
+++ b/packages/nextjs/hooks/scaffold-stark/__tests__/useScaffoldEventHistory.test.ts
@@ -22,7 +22,7 @@ vi.mock("@starknet-react/core", () => ({
 
 describe("useScaffoldEventHistory", () => {
   const mockContractName = "YourContract";
-  const mockEventName = "contracts::YourContract::YourContract::EventParser";
+  const mockEventName = "EventParser";
   const mockTargetNetwork = {
     id: "testnet",
     rpcUrls: {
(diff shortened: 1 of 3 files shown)


gianalarcon (changes requested):
Thanks for your contribution. Your solution is not working on my side, please make sure you update the scrip tests related to [useScaffoldEventHistory](https://github.com/Scaffold-Stark/scaffold-stark-2/blob/93849c0b755e67a532982edc10238acdf8b9d662/packages/nextjs/hooks/scaffold-stark/__tests__/useScaffoldEventHistory.test.ts#L2)

gianalarcon:
Testing your PR on ch-0 speedrun temporary [branch](https://github.com/Scaffold-Stark/speedrunstark/tree/ch-0/eventHistory). I added your PR on this [commit](https://github.com/Scaffold-Stark/speedrunstark/commit/d9e3642126d103e66244facdad405c895fb5c577)  
But got this error
Please also test the hook manually
image

metalboyrick (approved):
LGTM, sorry for the delay!

PR #567, opened 2025-06-18
chore: update deps

metalboyrick:
# Task name here

Updates snforge and starknet devnet to latest version


metalboyrick:
Closed due to unresolved dependencies on OpenZeppelin and Starknet Devnet

PR #586, opened 2025-06-30
feat(snfoundry): assert deployer is defined + improve error message

### LINKED ISSUE #570: Use ENV Validations on Contract Deployments
## Overview

Integrate an ENV validation system within the contracts section. Currently we get obscure error messages that might have been thought as a bug on scaffold, when in actuality it is just the deployer forgetting to fill in some ENV. (in particular the private keys and the address).

## Acceptance Criteria
- ~remove the need for inputting the address manually, we can derive them from the PK. We should not have to specify `ACCOUNT_ADDRESS_DEVNET`, `ACCOUNT_ADDRESS_SEPOLIA`, `ACCOUNT_ADDRESS_MAINNET`~
- Show proper color coded error messages if the private key is not set yet. We don't want obscure error messages

## Tips
- use https://starknetjs.com/docs/API/namespaces/ec.starkCurve/#getstarkkey to derive public keys from private keys.
- encode the environment variables into a typescript object and perform schematic validations there, this is a cleaner approach.

red8paw:
# ENV validations on contract deployments

Fixes #570

## Types of change

- [ ] Feature
- [ ] Bug
- [X] Enhancement

## Comments (optional)

- Signing and verifying typed data would confirm the private key matches the smart contract account, but that’s probably overkill for this case.

## Screenshot

The error message in terminal

image

### DIFF
--- a/packages/snfoundry/scripts-ts/deploy-contract.ts
+++ b/packages/snfoundry/scripts-ts/deploy-contract.ts
@@ -319,6 +319,14 @@ const exportDeployments = () => {
   fs.writeFileSync(networkPath, JSON.stringify(deployments, null, 2));
 };
 
+const assertDeployerDefined = () => {
+  if (!deployer) {
+    const errorMessage = `Deployment failed: deployer account is not defined - missing environment variables.\nPlease check your \`.env\` file for \`ACCOUNT_ADDRESS_${networkName.toUpperCase()}\` and \`PRIVATE_KEY_${networkName.toUpperCase()}\`.`;
+    console.error(red(errorMessage));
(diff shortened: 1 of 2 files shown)


red8paw:
> I think this should be ok no? the objective of this PR is to be able to log as many known errors properly.

✅ Agreed

> Do you think this will incur a significant performance hit?

✅ Pretty sure checking the signature on-chain doesn’t cause any performance issues
- already calling on-chain to check if the wallet is deployed
- readonly action

Should I include this in the PR?

red8paw:
> @red8paw yes please!

Updated
@metalboyrick Please take a look 🫡

PR #591, opened 2025-07-04
ci: deploy YourContract when deploy DemoApp

### LINKED ISSUE #574: Demo App Does Not Show `YourContract`
## Context 

The Demo App is supposed to show the `YourContract` tab but it is not there. We want to show it.

Image

## Acceptance Criteria

The Demo App should look like this:

Image

## Tips

Refer to https://github.com/Scaffold-Stark/scaffold-stark-2/issues/540 for starting points

red8paw:
# Show `YourContract` tab on Demo App

Fixes #574

## Types of change

- [ ] Feature
- [ ] Bug
- [X] Enhancement

## Comments (optional)
- Pre-deploy the contract on Sepolia first.
- If a developer deploys the contract again, it will overwrite the default config in the `deployedContracts.ts` file 🛠️

## Screenshot
image

### DIFF
--- a/.github/workflows/demo.yaml
+++ b/.github/workflows/demo.yaml
@@ -55,10 +55,11 @@ jobs:
           VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
           VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
           VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
+          RPC_URL_SEPOLIA: ${{ secrets.RPC_URL_SEPOLIA }}
         run: |
           cd source_repo
           vercel link --yes --project $VERCEL_PROJECT_ID --token $VERCEL_TOKEN --scope $VERCEL_ORG_ID
-          vercel --build-env NEXT_PUBLIC_IGNORE_BUILD_ERROR=true --build-env NEXT_PUBLIC_SEPOLIA_PROVIDER_URL=https://starknet-sepolia.blastapi.io/64168c77-3fa5-4e1e-9fe4-41675d212522/rpc/v0_8 --prod --token $VERCEL_TOKEN --scope $VERCEL_ORG_ID
+          vercel --build-env NEXT_PUBLIC_IGNORE_BUILD_ERROR=true --build-env NEXT_PUBLIC_SEPOLIA_PROVIDER_URL=$RPC_URL_SEPOLIA --prod --token $VERCEL_TOKEN --scope $VERCEL_ORG_ID
 
       - name: Notify Slack on Success
         if: success()
--- a/.github/workflows/demo.yaml
+++ b/.github/workflows/demo.yaml
@@ -31,6 +31,23 @@ jobs:
           node-version: 22
           registry-url: "https://registry.yarnpkg.com"
 
+      # TODO: setup yarn cache
+      - name: Install dependencies
+        run: |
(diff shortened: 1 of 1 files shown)


metalboyrick (changes requested):
GM @red8paw thanks for the PR.

I think better if we include it within the `predeployedContracts.ts` file, or we can have a CI that pre-deploys this when pushing to main

thoughts? @gianalarcon

gianalarcon:
Good finding @red8paw 
I think what we can do, is to run "yarn compile or deploy" after we set chain to sepolia, somewhere [here](https://github.com/Scaffold-Stark/scaffold-stark-2/blob/a0c511a0ba8b76d31f42fb102f8408ead3185d8e/.github/workflows/demo.yaml#L25)

red8paw:
> Good finding @red8paw
I think what we can do, is to run "yarn compile or deploy" after we set chain to sepolia, somewhere [here](https://github.com/Scaffold-Stark/scaffold-stark-2/blob/a0c511a0ba8b76d31f42fb102f8408ead3185d8e/.github/workflows/demo.yaml#L25)

I don’t think redeploying the contract every time we deploy the front-end is a good practice. Reasons:
- Random network issues can occur during contract deployment
- Adds unnecessary config for deployer address, gas settings, etc
- Our goal is to have basic contract interaction working on the Demo page

red8paw:
>I think better if we include it within the predeployedContracts.ts file, or we can have a CI that pre-deploys this when pushing to main

me too 👍 
I tried adding it to `predeployedContracts.ts`, but the `deepMergeConfig` function is handling the ABI array incorrectly.

Do you want me to look into fixing this as well?

red8paw:
> The demo interaction should be linked with the latest version of the sample contract on scaffold stark

Totally get it 🙏
But for me, if we just want to update the demo, running `yarn deploy --network sepolia` and create PR feels like a reasonable trade-off.

@metalboyrick let me know your call

I’ll update the PR accordingly

red8paw:
@gianalarcon @metalboyrick
I’ve updated the demo.yaml based on your feedback. appreciate a review when you have a moment 🙏

btw, these three ENVs should be set in GitHub Secrets:
- `PRIVATE_KEY_SEPOLIA`
- `RPC_URL_SEPOLIA`
- `ACCOUNT_ADDRESS_SEPOLIA`

.github/workflows/demo.yaml
```diff
@@ -40,7 +57,6 @@ jobs:
           VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
         run: |
           cd source_repo
-          yarn install
           vercel link --yes --project $VERCEL_PROJECT_ID --token $VERCEL_TOKEN --scope $VERCEL_ORG_ID
           vercel --build-env NEXT_PUBLIC_IGNORE_BUILD_ERROR=true --build-env NEXT_PUBLIC_SEPOLIA_PROVIDER_URL=https://starknet-sepolia.blastapi.io/64168c77-3fa5-4e1e-9fe4-41675d212522/rpc/v0_8 --prod --token $VERCEL_TOKEN --scope $VERCEL_ORG_ID
```
gianalarcon, line 61: here we using a public sepolia rpc url, we can replace the hardcode with `secrets.RPC_URL_SEPOLIA` as well

metalboyrick:
I have added the `RPC_URL_SEPOLIA` secret into the secrets

metalboyrick:
Added the other two secrets

PR #621, opened 2025-07-24
fix: handle missing contructor argument

### LINKED ISSUE #601: [SS2] Optimize Missing Constructor Error in Deployment
## Context

Currently we have an issue where that does not indicate that some constructor values are missing when we are deploying a new contract. We would like to fix this by providing an explicit message if this does happen

## Acceptance criteria

- When someone forgots to input a constructor when deploying contracts, we should indicate that they did not put a constructor in.
- We also need to be specific on what value is missing, or if the value is of wrong type.

od-hunter:
#Task name here

Fixes #601 

## Types of change

- [ ] Feature
- [ ] Bug
- [x] Enhancement

### DIFF
--- a/.github/workflows/demo.yaml
+++ b/.github/workflows/demo.yaml
@@ -19,6 +19,17 @@ jobs:
           token: ${{ secrets.ORG_GITHUB_TOKEN }}
           path: source_repo
 
+      - name: Install scarb
+        uses: software-mansion/setup-scarb@v1
+        with:
+          tool-versions: ./.tool-versions
+          scarb-lock: ./packages/snfoundry/contracts/Scarb.lock
+
+      - name: Install snfoundryup
+        uses: foundry-rs/setup-snfoundry@v3
+        with:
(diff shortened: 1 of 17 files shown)


metalboyrick (changes requested):
When I tried, it does not show a specific error:

The code:

image

The Logs:

metalboyrick (changes requested):
When I tried, it did not work:

Here is my deploy script:

```typescript
const deployScript = async (): Promise<void> => {
  await deployContract({
    contract: "YourContract",
    constructorArgs: {
      owner: "0x0",
    },
  });
};
```

Here is the results, no specific errors:

image

od-hunter:
That was because you tried it with the Zero address, i have added checks for zero address

metalboyrick:
yep, should be able to handle invalid args as well, i was doing it on purpose. Your messages on the logs also say "invalid ..."

od-hunter:
Hi @metalboyrick , please review.

PR #627, opened 2025-07-28
build: ci for checking deploy scripts

metalboyrick:
Currently the PR Verifier checks do not reach the scripts on the `snfoundry` directory, This PR adds those directory to the scope

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,6 +12,8 @@ on:
     paths:
       - "packages/nextjs/**"
       - "packages/snfoundry/contracts/**"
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -12,6 +12,8 @@ on:
     paths:
       - "packages/nextjs/**"
       - "packages/snfoundry/contracts/**"
```
gianalarcon, line 14: what about we just keep "packages/snfoundry/**"

gianalarcon (approved):

PR #634, opened 2025-08-04
chore(release): 1.1.0 [skip ci]

metalboyrick:
Release for 1.1.0


metalboyrick:
Please review first, I will merge when other dependencies are ok

package.json
0xquantum3labs, line 1: this will automatically be done when you close the PR and the merge commit has this `[minor]` at the end

packages/nextjs/services/web3/provider.ts
```diff
@@ -16,8 +16,40 @@ const containsDevnet = (networks: readonly chains.Chain[]) => {
...
+  let rpcUrl = "";
+
+  switch (networkName) {
+    case "devnet":
+      rpcUrl = devnetRpcUrl || fallBack || "http://127.0.0.1:5050";
+      break;
+    case "sepolia":
+      rpcUrl =
+        sepoliaRpcUrl ||
+        fallBack ||
```
0xquantum3labs, line 34: `NEXT_PUBLIC_PROVIDER_URL=http://127.0.0.1:5050`

The `.env.example` has this line, i think it would fall under `fallback` ALWAYS after a postinstall thus fail if i do `install` then change to sepolia. 

Overall this logic can have some complex scenarios

PR #718, opened 2026-07-20
chore: add portable drift-check workflow, harden .claude leakage paths

tu11aa:
## What

Adds `.claude/workflows/stark-drift-check.mjs` — a reusable maintenance workflow that audits this repo and its sync targets for toolchain drift, dependency lag, sync-workflow health, and template breakage (via a local devnet scaffold+deploy smoke test).

Also hardens the paths by which anything under `.claude/` could escape this repo.

## Why the hardening

This repo is **public with 171 forks**, and its workflows rsync the tree outward:

| Path | Destination |
|---|---|
| `sync-speedrun-repo.yaml:109` | speedrunstark (full tree) |
| `sync-basecamp-repo.yaml:114` | basecamp (full tree) |
| `release-create-stark.yaml:82` | `create-stark/templates/base` → **`npm publish`** |

The last one is the widest: anything committed here can reach `templates/base`, get published to npm, and land on the disk of every user running `npx create-stark`.

## Changes

**1. Portable workflow** — no machine-specific values. Agents derive repo root via `git rev-parse --show-toplevel` and org via `git remote get-url origin`; siblings are referenced as `owner/repo` through `gh api` rather than local clone paths. Verified: `grep -c '/Users/'` → 0.

**2. `.gitignore`** — the load-bearing change:
```
.claude/*
!.claude/workflows/
!.claude/skills/
.worktrees/
.gitnexus/
target/
```
Must be `.claude/*` (glob), not `.claude/` — git cannot re-include a path inside an ignored directory, so a bare `.claude/` would silently make the negations a no-op. A comment in the file records this. `.claude/skills/` is tracked alongside `.claude/workflows/` for the same reason — both are repo automation intended to ship, not local scratch state — and the two `stark-update-phase1`/`stark-sync-phase2` skill files already on disk are committed here.

`.worktrees/` was found live during this work: 1.5 GB of untracked nested checkouts at the repo root, previously unignored.

**3. `--exclude='.claude/'` on the rsyncs** — defense-in-depth, not the primary guard. CI obtains `source_repo` via `actions/checkout`, a fresh clone of **tracked files only**, so an untracked file never reaches rsync in the first place. The exclude covers `git add -f` or a future change that broadens an rsync's source. The one in `sync-rn-repo.yaml` is a documented no-op today (its source is `packages/snfoundry/`, not the repo root) and is kept deliberately for that same reason.

## Verification

```
grep -c q3labsadmin  → 0
grep -c /Users/      → 0
git add --dry-run .claude/workflows/probe.mjs  → add '...'        (addable)
git add --dry-run .claude/probe.json           → ignored          (blocked)
```

Note `git check-ignore` was deliberately **not** used as the acceptance test. It answers
*"which pattern matched this string"*, which is not the same question as *"will git ignore
this file"*. The two diverge along four axes, all reproduced on this repo (git 2.51.0):

| Query | Result |
|---|---|
| `.claude/workflows` (dir, contains tracked files) | exit 1, empty — index is consulted by default |
| `.claude/workflows` (dir, `--no-index`) | exit 0, matches `!.claude/workflows/` — **negation match still exits 0** |
| `.claude/nonexistent` (dir, not on disk) | exit 0, matches `.claude/*` — **reports ignored for a path that would not be** |
| `.claude/workflows/file.mjs` (file) | exit 1, empty — the correct answer |

The traps: trailing-slash rules like `!.claude/workflows/` only match *directories*, and git can
only know a path is a directory if it exists on disk — so a query for a non-existent path falls
through to the exclude rule and reports the exact opposite of the real behaviour. Separately,
`check-ignore` consults the index unless `--no-index` is passed, meaning its answer depends on
tracking state, which is often the very thing you are trying to establish.

Note variable (3) is **invisible if you query a file**: for a tracked *file*, both the default
and `--no-index` return exit 1/empty. It only surfaces on the *directory* query. Anyone
reproducing the table above with a file path will wrongly conclude it does not exist.

Use `git add --dry-run <file>` instead — it reports what git will actually do, and is immune to
all four variables. **But probe with a fresh, never-committed filename**, because it has one
failure mode of its own:

| Target | Output |
|---|---|
| untracked, not ignored | `add '.claude/workflows/__probe'` |
| untracked, ignored | `The following paths are ignored by one of your .gitignore files` |
| **already tracked** | **empty, exit 0** — reads as "nothing happened" and proves nothing |

So verifying against a file already committed (e.g. `stark-drift-check.mjs` itself) returns
silence regardless of the rules. The acceptance test used here:

```sh
echo x > .claude/workflows/__probe && git add --dry-run .claude/workflows/__probe
#   -> add '.claude/workflows/__probe'                    (addable, correct)
echo x > .claude/__probe          && git add --dry-run .claude/__probe
#   -> The following paths are ignored by one of your...  (blocked, correct)
rm -f .claude/workflows/__probe .claude/__probe
```

Always query a **file inside** the directory, never the directory itself, and never one that is
already tracked. Empty output means the path is already tracked — re-run with a name that does
not exist yet.

## Not in scope

Does not fix any drift the workflow found (the stale `v3-bulletproof-contracts` sync, the rn `verify-contracts.ts` import, the release→fork sync trigger). Those are separate PRs.

### DIFF
--- a/.claude/skills/stark-sync-phase2/SKILL.md
+++ b/.claude/skills/stark-sync-phase2/SKILL.md
@@ -1,0 +1,152 @@
+---
+name: stark-sync-phase2
+description: Phase 2 — propagate an already-merged scaffold-stark-2 update out to its sibling forks and the create-stark npm template. Verifies whether each target's sync automation actually worked, and repairs broken automation rather than hand-patching the target. Use only AFTER Phase 1 changes are merged; run stark-update-phase1 first.
+---
+
+# Phase 2 — Propagate to siblings
+
+Answer: **did each sibling actually receive the Phase 1 changes, and if not, why?**
+
+## Precondition — do not start without this
+
+Phase 2 propagates *already-merged* work. Verify Phase 1 actually landed before touching anything:
+
+```sh
+git log --oneline origin/main -5
+git log --oneline origin/main..origin/develop | wc -l    # unreleased work sitting on develop
+```
+
+If the Phase 1 changes are still on `develop` or still in an open PR, **stop**. Syncing now propagates a half-state, and re-syncing later produces conflicts on top of it.
+
+## Governing rule — repair, do not bypass
+
+When a target is out of date **and** has automation, the automation is the defect. Fix the workflow so it works, then let it run.
+
+Do **not** hand-patch the target repo. A manual edit gets overwritten by the next successful sync, hides the broken automation, and has to be redone every release. The one exception is a target with **no** automation at all — see the routing table.
+
+---
+
+## Targets
+
+| Target | Kind | Automation | Trigger |
+|---|---|---|---|
+| `Scaffold-Stark/speedrunstark` | fork | `sync-speedrun-repo.yaml` | push(develop) + PR-closed + **dispatch** |
+| `Scaffold-Stark/basecamp` | fork | `sync-basecamp-repo.yaml` | push(develop) + PR-closed + **dispatch** |
+| `Scaffold-Stark/scaffold-stark-rn` | fork (sed-rewritten) | `sync-rn-repo.yaml` | push(develop) + PR-closed + **dispatch** |
+| `v3-bulletproof-contracts` | **branch inside this repo** | `sync-bulletproof-contracts.yaml` | push(develop) **only** |
+| `Scaffold-Stark/create-stark` | **npm package template** | `release-create-stark.yaml` | PR-closed **only** |
+
+Two structural constraints that shape everything below:
+
+- **`sync-bulletproof-contracts` and `release-create-stark` have no `workflow_dispatch`.** They cannot be triggered manually. If they need to run, either a qualifying event must occur or the workflow needs a `workflow_dispatch` block added first.
+- **Releases are pushed to `main` as `chore(release): X` with `[skip ci]`.** Sync workflows fire on `push: branches:[develop]` or `pull_request: closed`. A release push therefore fires **nothing**. Do not assume a release propagated.
+
+---
+
+## Step 1 — Did the automation run, and did it *work*?
+
+These are different questions. Run all three checks per target; a target passes only if all three agree.
+
+```sh
+# (a) did it run at all?
+gh run list --workflow <file> --limit 10 --json conclusion,createdAt,headBranch
+
+# (b) does the trigger still resolve? (a deleted branch silently kills a workflow)
+git ls-remote --heads origin develop
+
+# (c) DID IT PRODUCE ANYTHING? — the check that actually matters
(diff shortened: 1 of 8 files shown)


metalboyrick (changes requested):
Reviewed via adversarial multi-agent pass (CI/workflow mechanics + a separate leakage threat-model), with the git behavior reproduced locally rather than argued from the docs. The rsync hardening is correct and the `.claude` root leak is genuinely closed. One should-fix on the `.gitignore`, plus two low-value disclosures to consider.

**Verified good**
- All four rsync `--exclude='.claude/'` additions are placed before the include rules, so `.claude/` cannot reach the sync forks or the npm `templates/base`. `--delete` does not wipe an existing `.claude/` in a destination fork. The `sync-rn-repo.yaml` exclude is a genuine no-op today (its source is `packages/snfoundry/`) and is correctly documented as defense-in-depth.
- `.claude/settings.local.json`, `.claude/settings.json`, and `.claude/scratch/*` all stay ignored under the new glob. Confirmed with real git.

**Should-fix: the negations re-include the whole subtrees, not just the payload**
`.gitignore` — `!.claude/workflows/` and `!.claude/skills/` re-include the entire subtrees, not only the intended `.mjs` + `SKILL.md` files. Reproduced on git 2.43.0 and again here: with this exact `.gitignore`, `git add -A` stages `.claude/skills/.env`, `.claude/workflows/settings.local.json`, and `.claude/skills/stark-update-phase1/tmp/state.json`. These are the exact secret/scratch/settings shapes the `.claude/*` rule blocks one level up. Today's tracked files are clean, so this is latent, but the drift-check system is agent-run and its skills write state files, so a future misconfig writing into either directory would be committed to the public repo and cloned into all 171 forks with no per-file gate.
Fix (verified to ship only the three intended files while keeping the decoys ignored):
```
.claude/*
!.claude/workflows/
.claude/workflows/*
!.claude/workflows/*.mjs
!.claude/skills/
.claude/skills/*
!.claude/skills/*/
.claude/skills/*/*
!.claude/skills/*/SKILL.md
```

**Consider: this PR ships some internal posture into public files**
- `.claude/skills/stark-update-phase1/SKILL.md` publishes the repo's Dependabot backlog counts and names the open advisory `GHSA-5xrq-8626-4rwp`. Derivable from the public lockfile via `npm audit`, so low value, but it hands it over directly.
- `.claude/skills/stark-sync-phase2/SKILL.md` states which sync workflow fails with no Slack alerting, plus a specific failed run id. Also observable from the public `.github/workflows/`, so low value.
If these are intended as public template runbooks, they are fine to ship. If they are internal ops notes, keep them untracked.

**Merge order**
This PR and #720 both add the same `.claude/*` block to `.gitignore` (add/add conflict, trivial to resolve). More importantly, this PR is what adds the rsync excludes, so it should land before #720. Merging #720 first, while develop's rsyncs still have no `.claude/` exclude, would propagate #720's `.claude/` tree to the sync forks and npm.

Requesting changes for the `.gitignore` allowlist.

PR #719, opened 2026-07-21
chore(deps): bump vitest to patched 3.2.7, constrain daisyui to ^5.7.0

tu11aa:
Note that the vitest advisory GHSA-5xrq-8626-4rwp is dev-only (triggers only when the Vitest UI server is running, not in the deployed bundle) but the fix is free because the range already allowed it.

Corrections from the original description: the lockfile resolves vitest to **3.2.7** (a superset of the 3.2.6 fix), not 3.2.6. daisyui is constrained to a caret range (`^5.7.0`), not pinned to an exact version — also fine for a template repo.

This PR also brings `@vitest/coverage-v8` and `@vitest/coverage-istanbul` up to 3.2.7 in the lockfile. They were stuck at 3.2.4, a peer-version mismatch against the 3.2.7 vitest core that `yarn coverage` runs against. No `package.json` ranges changed — the existing `^3.0.9` range already permitted 3.2.7; this is a lockfile-only fix.

Verified:
- `yarn install --immutable` passes with no vitest-related peer warnings
- `yarn test` and `yarn coverage` both pass in `packages/nextjs`

### DIFF
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -71,7 +71,7 @@
     "@vitejs/plugin-react": "^4.3.4",
     "@vitest/coverage-istanbul": "^3.0.9",
     "@vitest/coverage-v8": "^3.0.9",
-    "daisyui": "latest",
+    "daisyui": "^5.7.0",
     "eslint": "^8",
     "eslint-config-next": "15.2.3",
     "jsdom": "^26.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3833,7 +3833,7 @@ __metadata:
     "@vitest/coverage-v8": ^3.0.9
     abi-wan-kanabi: ^2.2.4
     blo: ^1.1.1
-    daisyui: latest
+    daisyui: ^5.7.0
     eslint: ^8
     eslint-config-next: 15.2.3
     ethers: ^6.12.0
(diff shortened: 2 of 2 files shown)


metalboyrick (changes requested):
Reviewed via adversarial multi-agent pass (correctness + supply-chain), every fact web-verified against the npm registry and the GitHub advisory rather than the PR body. The bump is safe and the advisory is genuinely resolved. Two items before merge, one of them a real version skew.

**Verified good**
- `GHSA-5xrq-8626-4rwp` = CVE-2026-47429, Critical (CVSS 9.8), missing authorization in the Vitest UI/API dev server. Vulnerable `<3.2.6`, patched `3.2.6`. The lockfile ships `3.2.7`, which contains the fix. The "dev-only, not in the deployed bundle" claim is accurate: vitest is a devDependency and never enters the Next.js production build.
- The 52 changed `yarn.lock` lines are all vitest-ecosystem + daisyui. No unrelated package bumped, no new major, no registry/integrity change. `yarn install --immutable` passes.

**Should-fix: coverage plugins left behind, peer skew**
`packages/nextjs/yarn.lock` — vitest core and its internal `@vitest/*` deps move to `3.2.7`, but `@vitest/coverage-v8` and `@vitest/coverage-istanbul` stay at `3.2.4`. Both declare an exact peer `vitest: 3.2.4`. Since `packages/nextjs/package.json` has both coverage plugins as direct devDeps and a `"coverage": "vitest run --coverage"` script, `yarn coverage` now runs 3.2.4 coverage plugins against a 3.2.7 core. Coverage plugins hook vitest internals, which is why the peer is exact-pinned. It does not break `--frozen-lockfile` and CI (`vitest run --passWithNoTests`) is green, so this is not a merge blocker, but please align them.
Fix: bump `@vitest/coverage-v8` and `@vitest/coverage-istanbul` to `3.2.7` (the `^3.0.9` range already permits it).

**Cosmetic: title/body**
The title says "bump vitest to patched 3.2.6" but the lockfile ships `3.2.7`, and "pin daisyui" is actually a caret range `^5.7.0`, not a pin. Both outcomes are fine (3.2.7 is a superset of the fix; a caret is correct for a template repo). Please correct the wording so a reviewer trusting the title is not surprised.

Requesting changes only for the coverage-plugin skew. The title fix is optional.

PR #720, opened 2026-07-21
feat(.claude): drift-check + 14-step smoke-test gate for the Phase 1→2 pipeline, with fork/npm leakage hardening

tu11aa:
Supersedes and closes #718 and #725 — the three were folded into one PR because they are one concern and because splitting them created a merge-order hazard (see below).

This adds the agent tooling that sits around the Phase 1 → Phase 2 dependency pipeline, plus the hardening that keeps that tooling from escaping this repo into 171 forks and onto npm.

## Why one PR instead of three

The original review required strict order **#718 → #720**: #718 adds `--exclude='.claude/'` to the sync-fork rsyncs, and #720 adds the `.claude/` tree. Merging #720 first would have propagated that tree to the sync forks and to `create-stark/templates/base` → `npm publish`.

Folding them together removes the hazard rather than documenting it. There is no longer any intermediate `develop` state that has the content but not the excludes. #725 was already stacked on this branch and only ever needed to land with it.

## What's in it

**1. Drift-check workflow** (`.claude/workflows/stark-drift-check.mjs`) — audits this repo and its sync targets for toolchain drift, dependency lag, sync-workflow health, and template breakage. Portable: repo root via `git rev-parse --show-toplevel`, org via `git remote get-url origin`, siblings addressed as `owner/repo` through `gh api`. No machine-specific values.

**2. A 14-step end-to-end gate** (`.claude/skills/stark-smoke-test/`) that runs *after* Phase 1 updates merge and *before* anything propagates in Phase 2. It proves the stack works on devnet instead of assuming a green `yarn install` means a working app. **A red gate blocks Phase 2 propagation.**

| Steps | Owner | What |
| --- | --- | --- |
| 1–7 | `stark-smoke-test.mjs` | preflight (toolchain vs `.tool-versions`, ports, devnet `.env`), start devnet, cross-check predeployed accounts, deploy, start dev server |
| 8–13 | `SKILL.md` (agent-side) | Chrome over raw CDP: console/network clean, connect burner wallet, read a value, send one write tx, confirm it changed, screenshots |
| 14 | `stark-smoke-test.mjs` | `yarn test` and `yarn coverage` in `packages/nextjs` |

**3. An opt-in Sepolia deploy gate** (S1–S5) for the cases devnet cannot cover, kept deliberately out of the default path so live-RPC flakiness never produces a false red.

**4. Leakage hardening.** This repo is public with **171 forks**, and its workflows rsync the tree outward:

| Path | Destination |
|---|---|
| `sync-speedrun-repo.yaml` | speedrunstark (full tree) |
| `sync-basecamp-repo.yaml` | basecamp (full tree) |
| `release-create-stark.yaml` | `create-stark/templates/base` → **`npm publish`** |

The last is the widest — anything committed here can reach npm and land on the disk of every `npx create-stark` user. So: `--exclude='.claude/'` on all four rsyncs (defense-in-depth; CI's `actions/checkout` already ships tracked files only), plus a `.gitignore` allowlist that ships exactly the intended files and nothing else.

The `.gitignore` must use `.claude/*` (glob), not `.claude/` — git cannot re-include a path inside an ignored directory, so a bare `.claude/` would silently make every negation a no-op. A comment in the file records this.

## Design notes

- **Deterministic orchestration goes in the script, judgment goes to the agent.** Agents guess "it's probably up by now" and continue too early, producing false failures. "Is this page actually correct" is the part a script cannot answer.
- **`up` / `down` modes.** `up` detaches children so they outlive it and records PIDs to `.smoke-test-state.json`; `down` kills them and confirms ports released. Process identity (`ps -o lstart=`) is recorded at spawn and re-verified before any kill, so a reused PID after a crash+reboot cannot cause a stray group-kill.
- **On failure the stack stays up**, so the exact failure state is debuggable.
- **Never silently edits `.env`.** The devnet block ships commented out in `.env.example`, so step 3 failing is the expected first-run result. The script prints the block and *offers* to write it (`--write-env` / interactive confirm).
- **Exit 0 is not accepted as proof of deploy.** Step 6 also requires `deployedContracts.ts` mtime to have moved *and* a class hash to be present. The Sepolia S5 check goes further and validates freshness against *that run's own* deploy output, so a deploy that exits 0 without rewriting the file cannot confirm a previous deployment.

## Rejected alternatives

1. **Pure Markdown skill** — agents are bad at waiting for async dev servers to boot before opening browsers; flaky false negatives.
2. **Playwright committed to the repo** — this is a template repo consumed by `create-stark`; forcing Playwright deps and config onto every downstream fork is not acceptable. Plain Node ESM built-ins only.
3. **Sepolia in the default gate** — needs funded accounts and live RPCs; rate limits or an empty wallet produce false reds, which trains people to ignore the gate. Devnet by default, Sepolia opt-in.

## Review items addressed

All blocking and non-blocking items from the reviews on #718, #719-adjacent, and #720:

- `.gitignore` negations re-included whole subtrees — replaced with an explicit per-file allowlist, verified against decoy files (`.claude/skills/.env`, `.claude/workflows/settings.local.json`, `.claude/skills/*/tmp/state.json`) which stay ignored while the real `SKILL.md` + `.mjs` files stay tracked.
- Internal identifiers (`q3labsadmin`, a private repo path, an internal project name) shipped in `braavos.mjs` — stripped; `grep` for them now returns 0 repo-wide.
- Sepolia S1 misclassified a local toolchain problem as RED — now routed through `sepoliaInfra` (exit 2), matching the env-var half of the same step.
- The browser `waitFor()` could hang the gate forever if a Chrome target died mid-poll — all pending CDP requests now reject on ws `close`/`error`, and every `send` races a 30s timeout.
- `starknet_getClassHashAt` treated any JSON-RPC error as RED — now only `CONTRACT_NOT_FOUND` (20) is RED; anything else is INFRA.
- S5 freshness compared mismatched "firsts" on a multi-contract deploy — now matches against the deploy log's own address as ground truth, and the downstream class-hash confirmation uses that same address.

## Verification

Run on this branch, not asserted:

- `up` on a clean checkout stopped at **step 3** and refused to write `.env` without consent — the correct result.
- `up --write-env` reached **step 6** and failed on a corrupted local scarb cache, printing the step number and verbatim compiler output, leaving devnet alive with its PID and teardown command.
- After `scarb cache clean`, **steps 1–7 passed** against a real deploy (`YourContract`, class hash present), dev server answering `200` on :3000.
- Step 5 correctly matched `.env`'s unpadded `0x4b3f…` to devnet's `0x04b3f…` predeployed account #9 — the zero-padding normalisation this check needs.
- `down` stopped both processes and released 5050/3000; a second `down` exited 0 with no error.
- Full `run` passed all 14 steps end to end. Then, with `__test__/eventsData.test.ts` deliberately broken, `run` failed at **step 14/14** with the verbatim vitest assertion — confirming the step actually gates rather than reporting.
- Sepolia S1–S5 exercised against live Sepolia with on-chain confirmation of both the deploy and a write transaction.

## Not covered

Steps 8–13 are agent-side at runtime and are documented in `SKILL.md`, not executed by CI. CI does not run this gate — the `.claude/`-only diff does not trigger the existing workflow.


metalboyrick (changes requested):
Reviewed via adversarial multi-agent pass (security/secret-handling + script correctness), with the Starknet RPC facts web-verified and the key findings confirmed against the diff. This is careful work: the gate has no false-GREEN path, so it cannot pass a broken stack to Phase 2, which is the property that matters most. The devnet step 6 (mtime + class-hash) and the Sepolia S5 (freshness + non-zero class) proofs are both sound. No keys leak, there is no eval/shell-injection/remote-exec, and nothing auto-runs on install. The defects below are all in the false-RED / hang / info-disclosure direction.

Note on scope: this branch already contains #721 (Sepolia S1-S5) and the #724 refactor, so this review covers all of that.

**Should-fix**

1. Internal identifiers shipped publicly — `.claude/skills/stark-cdp-with-wallet/braavos.mjs` (3 lines). The comments carry `/Users/q3labsadmin/Q3/Brove/brove/packages/nextjs/e2e/helpers/braavos.ts`, disclosing a macOS username, a private repo path, and the internal project name "Brove" to the public repo and to every `create-stark` user. #718's own body claims `grep q3labsadmin -> 0`; this file reintroduces it. Strip the absolute paths and reference the helper generically.

2. Sepolia S1 misclassifies a local toolchain problem as RED — `stark-smoke-test.mjs:4344`. `sepoliaStepPreflight()` reuses `stepToolVersions()`, which on a `.tool-versions` mismatch or a missing binary calls `fail()` -> `process.exit(1)` = RED, blocking Phase 2. The three-way contract classifies an unconfigured/local condition as INFRA (exit 2). The env-var half of the same step already uses `sepoliaInfra()` (exit 2), so two "unconfigured" conditions in one step get opposite exit codes. The inline comment even documents the reuse. Route the toolchain check through `sepoliaInfra(...)` on the Sepolia path.

3. The browser `waitFor()` can hang the whole gate forever — `stark-smoke-test-browser.mjs`. Every step 8-13 probe awaits a `CdpSession` promise stored in `_pending`, resolved only by a matching CDP message. There is no `ws` `close`/`error` handler that rejects `_pending`, and `waitFor` only re-checks its deadline between iterations. If the Chrome target dies mid-poll (renderer crash, target closed during the step 11 tx wait), `await probe()` never settles, the deadline is never re-evaluated, and the gate hangs with no exit and no RED, stalling CI. You already fixed exactly this in `braavos.mjs` (`withTimeout`, "a target that goes unresponsive leaves the promise pending forever"); the main smoke-test probes have no equivalent guard. Race each `session.send` against a timer, or reject all `_pending` on ws close/error.

4. `.gitignore` negations re-include the whole `.claude/{workflows,skills}` subtrees (shared with #718). Reproduced: `git add -A` stages `.claude/skills/.env` and similar. See the allowlist fix in the #718 review. Please keep the `.gitignore` identical across both PRs.

**Consider**
- `stark-smoke-test.mjs:4506` — S5 treats any JSON-RPC error from `starknet_getClassHashAt` as RED "no class". After S2/S3/S4 already passed, a provider that returns a transient internal error as an HTTP-200 JSON-RPC error object gets classified RED and blocks Phase 2 even though the deploy may have landed. Switch on the code: `CONTRACT_NOT_FOUND` (20) is RED, everything else is INFRA.
- `down()`/`up()` trust bare PIDs from `.smoke-test-state.json` and send SIGKILL to the negative PID (whole group). After a crash + reboot + PID reuse, `alive(pid)` passes for an unrelated process and `down()` group-kills it. Record and re-verify a cheap process identity (start time or cmdline) before killing.
- S5 freshness compares the first "Contract Deployed at" address in the S4 log against the first address under the `sepolia:` key in `deployedContracts.ts`. For a multi-contract deploy those two "firsts" need not be the same contract, producing a false RED. Match by contract name, or against any address under the entry.
- The GIF/transcript redaction (`stark-demo-terminal.mjs`) is a weak secondary check: the GIF byte-grep cannot match a secret rendered into LZW-compressed pixels, and the transcript substring match misses a soft-wrapped secret. Not reachable today (deploy output is redacted on the only path that prints it), but please normalize whitespace before matching and stop logging "GIF bytes all clear" as if the frames were scanned.

**Merge order** — land #718 first so develop's rsyncs exclude `.claude/` before this PR adds its `.claude/` tree; otherwise this content propagates to the sync forks and npm `templates/base`.

Requesting changes for items 1-3 (and the shared `.gitignore`). The rest are non-blocking.

